PR #36, opened 2015-08-04
Add troubleshooting

brenoc:
After some time trying to figure out why my hot reloader was not working, I finally managed to find out what was wrong. This might be helpful for other people.


tamorim:
Nice! This will be very helpful :)

PR #66, opened 2015-10-28
Change prompt message to a more meaningful one

brenoc:
Change login prompt messages from this:
```
> account
> email
> password (typing will be hidden)
```
To:

```
> account name of the store you are developing
> email
> password (typing will be hidden)
```

### DIFF
--- a/src/lib/auth.coffee
+++ b/src/lib/auth.coffee
@@ -38,7 +38,10 @@ class AuthenticationService
     prompt.delimiter = ''
 
     prompt.start()
-    console.log 'Please log in with your VTEX credentials.', '\n'
+    console.log 'Please log in with your VTEX credentials:\n'.green +
+                'account  - The store account you want to be developing on\n' +
+                'login    - Your VTEX registered email\n' +
+                'password - Your VTEX registered password\n'
 
     prompt.get options, (err, result) =>
       if err then console.log '\nLogin failed. Please try again.'


tamorim:
I disagree with such a huge name, I prefer to inform what an account is via the docs, or find a 2 word name to be prompted.

dfosco:
I was confused with this detail the first time I ran the new toolbelt, and I'm sure many developers will overlook this detail in their _first interaction using it_.

Are you sure a small explanation isn't worth it? Just to save the quick trip to the docs. What about:

```
> account (name of the store you are developing)
> email
> password (typing will be hidden)
```

brenoc:
I agree with @dfosco.

I think is more intuitive and less error prone. Docs may say how it's done right, but won't make it more intuitive to use.

The dream of any UX is *"no docs needed"*

tamorim:
I didn't said that I don't agree, I think that it's a little confusing, indeed.
I just think that the message shouldn't as long, it should be something subtle.
That's why I said that something with 2 words is ideal.

tamorim:
`store account`?

dfosco:
@tamorim If the goal is clarity, I see no reason why we should keep to two words — `store account` is somewhat self-explanatory, but not the most intuitive way to phrase it.

tamorim:
Well, let me illustrate why I don't like it:

(actually it shows `login`, not `email`)
```
> account name of the store you are developing
> login
> password (typing will be hidden)
```

- I don't think a **terminal prompt** is a place for that
- What about email and password? Some people might get confused to what it means too
- It's a big arrow saying "Hey. we couldn't find a nice word to describe what we want, so we put his thing instead"

Maybe it's a better approach to do something like:

```
$ vtex login
account - The store account you want to be developing on
login - Your VTEX registered e-mail
password - Your VTEX registered password

> account
> login
> password (typing will be hidden)
```

I don't think it's perfect, but it's better because doesn't get in the way of the prompt and explains all prompts at once, so the user will understand everything he has to type beforehand.

What do you guys think?

dfosco:
@tamorim That seems like an ideal setup for me. Will the descriptions appear every time the user wants to login?

tamorim:
@dfosco Yes.

dfosco:
@tamorim We have a winner!!! :checkered_flag:

brenoc:
:+1:

tamorim:
@brenoc @dfosco Updated!

PR #68, opened 2015-10-28
Add more error messages to ease debugging

brenoc:
Doing this so we may debug #65


tamorim:
Fair enough, thanks @brenoc !

tamorim:
Just to leave noted, I think that this is more of a fix than a feature! That's why I released a patch.

PR #81, opened 2015-11-17
Fix access token prompt

### LINKED ISSUE #76: When logging in with access token password should not be hidden
When using the access token to log in, the input should not be hidden and the label should be "Access token" instead of "Password".

brenoc:
Fix #76


brenoc:
Deprecated in favor of es6

PR #83, opened 2015-11-18
Change codebase to ES6

brenoc:
Moving on!

PR #85, opened 2015-11-18
Show publish error details

brenoc:
Based of feature/es6


brenoc:
Depreacted in favor of #83

PR #142, opened 2016-06-10
Implement app's commands

tamorim:
Depends on https://github.com/vtex/apps-client-node/pull/2


src/modules/apps.js
```diff
@@ -7,40 +109,150 @@ export default {
     description: 'List your installed VTEX apps',
     handler: () => {
       log.debug('Starting to list apps')
-      log.info('You have no installed apps')
+      workspaceAppsClient().listDependencies(
+        getAccount(),
+        getDevWorkspace(getLogin()),
+        'storefront'
```
brenoc: Why filter by "storefront"?

src/modules/apps.js
```diff
@@ -7,40 +109,150 @@ export default {
...
     description: 'Install the specified app',
-    handler: (app) => {
+    handler: (app, options) => {
       log.debug('Starting to install app', app)
-      log.info('Install app', app)
+      const appRegex = new RegExp(`^${vendorPattern}\.${namePattern}@${wildVersionPattern}$`)
+      if (!appRegex.test(app)) {
+        return log.error('Invalid app format, please use <vendor>.<name>@<version>')
+      }
+      const simulation = options.s || options.simulation
```
brenoc: What a "simulation" does?

src/modules/apps.js
```diff
@@ -7,40 +109,150 @@ export default {
...
+        getAppManifest(root),
+        getVtexIgnore(root),
+      ])
+      .spread(({vendor, name, version}, ignore) => {
+        log.info('Watching app', `${vendor}.${name}@${version}`)
+        return all([
+          listFiles(root, ignore),
+          keepAppAlive({vendor, name, version}),
+        ])
+        .spread(localFiles => all([
```
brenoc: Where Render build step would come in here?

src/file.js
```diff
@@ -0,0 +1,159 @@
...
+    .on('ready', resolve)
+  })
+}
+
+export function sendSaveChanges (root, file, sendChanges) {
+  log.debug('Sending save changes', file)
+  return sendChanges(createChanges(root, { [file]: 'save' }))
+}
+
+export function sendRemoveChanges (root, file, sendChanges) {
```
brenoc: This file is incredibly big. I think we should split it. There's no way all this functions are cohesive.

brenoc:
@firstdoit @tamorim what about my questions? lol

src/modules/apps.js
tamorim: @brenoc Yeah, list only for storefront doesn't make sense.
BUT, a service is required. Should we make the service an argument?

src/modules/apps.js
tamorim: @brenoc It's like a dry run flag, it won't install or uninstall the apps

src/modules/apps.js
tamorim: @brenoc Probably after or in parallel with the `watch`

firstdoit:
@brenoc I didn't knew the answers 😬

src/modules/apps.js
firstdoit: why not call it dry run?

tamorim: @firstdoit We can rename it, i just named it after the API

PR #163, opened 2016-08-01
Fix vtex workspace link

firstdoit:
Update the link shown to the user during the watch command.

The query string key changed to `vtex_workspace` to match the cookie name. Also, we're publishing the new render only in `alpha` right now.

`vtex watch`, see the new link is issued.

image

#### Types of changes


tamorim:
@firstdoit LGTM :+1:

PR #190, opened 2016-08-23
Add commands to deal with app settings

rdumont:
It creates one new namespace (`settings`) with three new commands to deal with app settings, as follows.

```
settings <app> [field]                     Get app settings
settings set <app> <field> <value>         Set a value
settings unset <app> <field>               Unset a value
```

This enables developers to manage apps' settings without leaving the command line.

1. Get an app's settings.
2 . Set a value and get settings again. The new value should be listed.
3. Get only a field from the app's settings.
4. Unset the field, then get settings again. The field should be gone.

settings-usage

#### Types of changes
- [x] Requires change to documentation, which has been updated accordingly.


src/modules/apps.js
```diff
@@ -281,4 +282,57 @@ export default {
...
+    },
+
+    set: {
+      description: 'Set a value',
+      requiredArgs: ['app', 'field', 'value'],
+      handler: async (app, field, value) => {
+        try {
+          let jsonPath = '$.' + field
+          let patch = {}
+          jp.value(patch, jsonPath, value)
```
tamorim: @rdumont `path` cant receive the value from `jp.value`? This kind of mutation is quite awful...

tamorim: @rdumont Also, why wrap the whole code inside a try/catch? I think it would be better to wrap only where its needed.

tamorim:
@rdumont There's a lot of variable declarations using `let` that could be using `const` instead.

src/modules/apps.js
tamorim: @rdumont Don't forget you can use string interpolation using backticks and ${} inside.

rdumont: @tamorim, about the three issues:
1. No, the library doesn't support that :(
2. Sure, I was using that a lot to test, but they can be removed.
3. What line are you talking about?

tamorim: @rdumont on lines 296, 309 and 326.
But maybe you used the `+` operator for readability?

rdumont: Yes, I used it on purpose, since the string is a simple concatenation and contains a `$`. I find it more readable.

rdumont:
@tamorim, I fixed the issues you mentioned.

tamorim:
@rdumont LGTM :+1:

PR #197, opened 2016-09-08
Add init command

tamorim:
Add an `init` command that creates files and folders necessary for a VTEX app.

Friction on creating new apps.

Running `vtex init`.

n/a

#### Types of changes

### DIFF
--- a/src/modules/init.js
+++ b/src/modules/init.js
@@ -1,0 +1,144 @@
+import {join} from 'path'
+import chalk from 'chalk'
+import moment from 'moment'
+import log from '../logger'
+import inquirer from 'inquirer'
+import {mkdir, writeFile} from 'fs'
+import {Promise, promisify, mapSeries} from 'bluebird'
+
+const bbMkdir = promisify(mkdir)
+const bbWriteFile = promisify(writeFile)
+
+function promptService () {
+  return Promise.try(() =>
+    inquirer.prompt({
+      name: 'service',
+      message: 'Choose the VTEX service you will use',
+      type: 'list',
+      choices: ['render'],
+    })
+  )
+  .then(({service}) => service)
+}
+
+function promptName () {
+  const message = 'The app name should only contain number, letters, underscores and hyphen.'
+  return Promise.try(() =>
+    inquirer.prompt({
+      name: 'name',
+      message: 'What\'s your VTEX app name?',
+      validate: s => /^[A-Za-z0-9\-_]+$/.test(s) || message,
+      filter: s => s.trim(),
+    })
+  )
+  .then(({name}) => name)
+}
+
+function promptVendor () {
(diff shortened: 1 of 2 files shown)


src/modules/init.js
```diff
@@ -0,0 +1,144 @@
...
+  }
+  return JSON.stringify(manifest, null, 2)
+}
+
+export default {
+  init: {
+    handler: function () {
+      log.info('Hello! I will help you generate basic files and folder for your app.')
+      log.info('Please login before procceding.\n')
+      return this.login.handler()
```
firstdoit: Curiosity: why isn't this contemplated on the normal "login required" flow?

src/modules/init.js
```diff
@@ -0,0 +1,144 @@
...
+            join(process.cwd(), fullName, 'manifest.json'),
+            createManifest(name, vendor, title, description)
+          )
+        )
+        .then(promptService)
+        .then(service => this.init[service].handler(appFolder))
+        .tap(() => {
+          console.log('')
+          log.info(`${fullName} structure generated successfully!`)
+          log.info(`Enter on the ${fullName} folder and run ${chalk.bold.green('vtex watch')} to start developing!`)
```
firstdoit: `npm init` doesn't create a folder - instead, it grabs the current directory's name and suggests it as the package name. I think we should do the same so we maintain the same experience.

firstdoit:
LGTM pending two small details

src/modules/init.js
tamorim: Because we don't make any calls to the API that needs authentication.

src/modules/init.js
tamorim: @firstdoit Ok, I was struggling to choose if I should put on root or create a new folder.

src/modules/init.js
firstdoit: You don't need to receive an auth error to request login. If the command is not in this whitelist, it will automatically request login. https://github.com/vtex/toolbelt/blob/master/src/cli.js#L30
One other thing: What if the user is already logged in?

src/modules/init.js
firstdoit: Sure, it's more or less equivalent, but I'd rather follow `npm` in this case, right? :D

tamorim:
@firstdoit Done!

firstdoit:
⚡ LGTM 💥  good job :D

src/modules/init.js
tamorim: @firstdoit You're right! I forgot about the whitelist :/
Maybe it didn't prompt me because I was already logged.
I will remove this to take advantage of what's already there :)
Thanks for the reminder!

src/modules/init.js
tamorim: Sure, sure!

PR #209, opened 2016-09-29
Automatically add glob stars to folders in vtexignore

firstdoit:
Make our `.vtexignore` behave more similarly to `.gitignore`.

Because the way `node-glob` works, if you list `src/` in your `.vtexignore` all files inside will still be matched. To ignore everything, you need to actually force `src/**`. This adds stars automatically.
This is not the optimal solution, as we should conform to the actual [spec](https://git-scm.com/docs/gitignore), or maybe use something like https://github.com/kaelzhang/node-ignore, but it improves the experience for now.

Ignore a folder using `/` at the end, watch things not get sent.

N/A

#### Types of changes

### DIFF
--- a/src/modules/apps.js
+++ b/src/modules/apps.js
@@ -198,6 +198,7 @@ export default {
       return createTempPath(id).then(t => { tempPath = t })
       .tap(() => log.debug('Listing local files...'))
       .then(() => listLocalFiles(root))
+      .tap(files => log.debug('Compressing files:', '\n' + files.join('\n')))
       .then(files => compressFiles(files, tempPath))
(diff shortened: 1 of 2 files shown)


src/modules/apps.js
```diff
@@ -265,6 +266,7 @@ export default {
 
       return createTempPath(id).then(tempPath =>
         listLocalFiles(root)
+        .tap(files => log.debug('Sending files:', '\n' + files.join('\n')))
```
tamorim: @firstdoit Why `Sending files`? We're not actually sending any files right now.

firstdoit: We will in the following two lines 😜

PR #217, opened 2016-10-11
Fragment commands to leverage new findhelp support for lazy loading

tamorim:
Fragment the commands to compose a lazy loaded command tree.

Extensibility and performance.

Toolbelt should work as before, just slightly faster :smile: 

n/a

#### Types of changes


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "lib/index.js",
   "bin": "lib/cli.js",
   "scripts": {
-    "watch": "babel src --out-dir=lib --watch",
+    "watch": "rimraf lib/ && babel src --out-dir=lib --watch",
```
firstdoit: Use `run-s` instead of `&&` so we don't break on Windows ;)

package.json
```diff
@@ -56,7 +55,8 @@
     "eslint-config-vtex-react": "^1.0.2",
     "eslint-plugin-react": "^5.0.1",
     "eventsource": "^0.2.1",
-    "findhelp": "^0.1.4",
+    "fast-async": "^6.0.34",
```
firstdoit: 😍

tamorim:
@firstdoit Done!

PR #220, opened 2016-10-25
Filter empty files on watch

tamorim:
Prevent empty files from being sent to apps/vbase.

Resource not found when trying to get an empty file on vbase.

Trying to send any empty files during the watch startup or via changes.

n/a

#### Types of changes

### DIFF
--- a/src/file.js
+++ b/src/file.js
@@ -19,8 +19,9 @@ export function listLocalFiles (root) {
   })
   .then(files =>
     Promise.all(
-      files.map(file => bbStat(path.join(root, file))
-      .then(stats => ({file, stats})))
+      files.map(file =>
+        bbStat(path.join(root, file)).then(stats => ({file, stats}))
+      )
     )
   )
   .then(filesStats =>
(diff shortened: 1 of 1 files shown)


src/file.js
```diff
@@ -16,6 +17,17 @@ export function listLocalFiles (root) {
     nodir: true,
     ignore: getIgnoredPaths(root),
   })
+  .then(files =>
+    Promise.all(
+      files.map(file => bbStat(path.join(root, file))
+      .then(stats => ({file, stats})))
```
firstdoit: I've had to re-read this three times to get that `bbStat(path.join(root, file)).then(stats => ({file, stats}))` is one thing. Can I suggest you either keep it in the same line or rearrange it so the scope of `file` is clearer?

```
files.map(file =>
  bbStat(path.join(root, file))
  .then(stats => ({file, stats}))
)
```

src/file.js
```diff
@@ -16,6 +17,17 @@ export function listLocalFiles (root) {
...
+    )
+  )
+  .then(filesStats =>
+    filesStats.reduce((acc, {file, stats}) => {
+      return stats.size > 0 ? [...acc, file] : acc
```
firstdoit: Depending on how this is compiled by babel, it might get very inefficient. Might be better to `push` and `return` to avoid creating arrays. But *maybe* it does that already. :shrug:

PR #224, opened 2016-11-08
Allow batch install and uninstall (and uninstall without args)

tamorim:
To allow `install` and `uninstall` with multiple apps and make `uninstall` require an optional argument!
If you run `vtex uninstall` it will uninstall the app on the current folder.

Not having to type a lot to install or uninstall stuff :smile:  

Try to install and uninstall more than one app.

`vtex install vtex.shelf@0.5.3 vtex.react-slick@0.15.3`
`vtex uninstall vtex.shelf vtex.react-slick`

#### Types of changes

### DIFF
--- a/src/modules/apps/install.js
+++ b/src/modules/apps/install.js
@@ -8,33 +8,44 @@ import {
   wildVersionPattern,
 } from '../../manifest'
 
+const ARGS_START_INDEX = 2
+
+function installApps (apps) {
+  const app = apps.shift() || `${manifest.vendor}.${manifest.name}@${manifest.version}`
+  log.debug('Starting to install app', app)
+  const appRegex = new RegExp(`^${vendorPattern}.${namePattern}@${wildVersionPattern}$`)
+  if (!appRegex.test(app)) {
+    log.error('Invalid app format, please use <vendor>.<name>@<version>')
+    return Promise.resolve()
+  }
+
+  return installApp(app)
+  .then(() => log.info(`Installed app ${app} successfully`))
+  .then(() =>
+    apps.length > 0
+      ? installApps(apps)
+      : Promise.resolve()
+  )
+  .catch(err => {
+    if (err.statusCode === 409) {
+      return log.error(`App ${app} already installed`)
+    }
+    return Promise.reject(err)
+  })
+}
+
 export default {
   optionalArgs: 'app',
   description: 'Install an app on the current directory or a specified one',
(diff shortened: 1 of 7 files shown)


src/modules/apps/install.js
```diff
@@ -8,33 +8,44 @@ import {
   wildVersionPattern,
 } from '../../manifest'
 
+const ARGS_START_INDEX = 2
+
+function installApps (apps) {
```
firstdoit: What if `apps` is `undefined`? I'd suggest being defensive: `apps = []`.

src/modules/apps/install.js
```diff
@@ -8,33 +8,44 @@ import {
...
-
-    return installApp(app)
-    .then(() => log.info(`Installed app ${app} successfully`))
-    .catch(err => {
-      if (err.statusCode === 409) {
-        return log.error(`App ${app} already installed`)
-      }
-      return Promise.reject(err)
-    })
+    const apps = options._ ? [optionalApp, ...options._.slice(ARGS_START_INDEX)] : [optionalApp]
```
firstdoit: Assuming `options._` exists. Not necessarily true! I'd rather receive `...apps, options`, no? Does that even work? 😁

src/modules/apps/install.js
```diff
@@ -8,33 +8,44 @@ import {
...
+  .then(() =>
+    apps.length > 0
+      ? installApps(apps)
+      : Promise.resolve()
+  )
+  .catch(err => {
+    if (err.statusCode === 409) {
+      return log.error(`App ${app} already installed`)
+    }
+    return Promise.reject(err)
```
rdumont: Let's say I ask for `vtex install foo bar zaz`, then `foo` succeeds, but `bar` fails. This way, `zaz` would be ignored. I think that either `zaz` should be installed anyway, or at least the toolbelt should warn that it was not installed due to previous errors.

Something like:
```sh
$ vtex install foo bar zas
info:    foo installed
error:   could not install bar because ...
warn:    skipped installation of zaz due to previous errors
```

tamorim: @rdumont I agree! I think the notification is a better way.
I think this is optimal because if you're installing the dependencies first and then the dependents, it would break to go on without installing the dependencies first.

tamorim: Actually, here's how a fail would look like
selection_005

I believe it's clear enough already.
What do you think?

PR #225, opened 2016-11-09
Add loaders for install and uninstall commands

tamorim:
When you install or uninstall an app the VTEX Apps response is just a part of what actually happens on the background.

If you install a render app, for example, after the VTEX Apps response the render service still have to pick up the installed app and build a bundle with it.

If you install and switch directly to the page, you won't see what you wanted because render is still working it's magic.

This PR aims to fix that by adding loaders and connecting with the courier service to receive the messages that other systems may pass when working with the install or uninstall of an app.

install/uninstall commands being too simple on it's success.

Just try to install or uninstall any app on your workspace :smiley: 

n/a

#### Types of changes

### DIFF
--- a/src/courier.js
+++ b/src/courier.js
@@ -5,9 +5,9 @@ import {timeStop} from './time'
 import endpoint from './endpoint'
 import stripAnsi from 'strip-ansi'
 import {manifest} from './manifest'
+import {clearAbove} from './terminal'
 import EventSource from 'eventsource'
 import {consumeChangeLog} from './apps'
-import {clearLine, cursorTo} from 'readline'
 import {locatorByMajor} from './locator'
 import {__, any, contains, map} from 'ramda'
 import {setSpinnerText, stopSpinner, isSpinnerActive} from './spinner'
@@ -22,11 +22,6 @@ const levelFormat = {
 }
 
 let timeoutId
-
-const clearAbove = () => {
-  clearLine(process.stdout, 0)
-  cursorTo(process.stdout, 0)
-}
 
 const stopAndLog = (log) => {
   const timeEnd = timeStop()
--- a/src/courier.js
+++ b/src/courier.js
@@ -8,6 +8,8 @@ import {manifest} from './manifest'
 import EventSource from 'eventsource'
 import {consumeChangeLog} from './apps'
 import {clearLine, cursorTo} from 'readline'
+import {locatorByMajor} from './locator'
+import {__, any, contains, map} from 'ramda'
 import {setSpinnerText, stopSpinner, isSpinnerActive} from './spinner'
 
 const courierHost = endpoint('courier')
@@ -18,6 +20,8 @@ const levelFormat = {
   warning: log.warn,
   error: log.error,
 }
+
+let timeoutId
 
 const clearAbove = () => {
   clearLine(process.stdout, 0)
@@ -44,9 +48,9 @@ const stopAndLog = (log) => {
   console.log(timedLog)
 }
 
-const logToConsole = (level, origin, message) => {
+const logToConsole = (level, origin, message, timeout, action) => {
   const isLogLevelInfo = log.level === 'info'
-  const {message: text, timeout} = typeof message === 'string'
+  const {message: text, timeout: msgTimeout} = typeof message === 'string'
     ? {message} : message
   if (isLogLevelInfo && isSpinnerActive()) {
     if (level === 'error') {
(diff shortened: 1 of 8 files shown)


src/modules/apps/uninstall.js
```diff
@@ -14,8 +18,23 @@ const promptAppUninstall = (app) => {
...
-  const app = apps.shift() || `${manifest.vendor}.${manifest.name}`
+function courierCallback (apps) {
+  let counter = 0
+  return () => {
+    clearLine(process.stdout, 0)
+    cursorTo(process.stdout, 0)
+    const app = apps[counter]
+    counter += 1
+    log.info(`Uninstaled app ${app} successfully`)
+    if (counter === apps.length) {
```
firstdoit: I understand why this is here right now, but I'd love it to be abstracted away soon. Very much outside of the interest of the `uninstall` module. Unfortunately, no easy way for `courier` to handle it rn.

src/modules/apps/uninstall.js
```diff
@@ -14,8 +18,23 @@ const promptAppUninstall = (app) => {
...
+      process.exit()
+    }
+  }
+}
+
+function uninstallApps (apps, preConfirm) {
+  const app = head(apps)
+  const decApp = tail(apps)
```
firstdoit: What is `dec`?

src/courier.js
```diff
@@ -63,38 +71,67 @@ const logToConsole = (level, origin, message) => {
...
+
+const originMatch = (msgOrigin, origin = []) => {
+  const msgOriginByMajor = appByMajor(msgOrigin)
+  if (origin.length === 0) {
+    origin.push(getOriginByManifest())
+  }
+  return any(app => msgOriginByMajor.indexOf(app) > -1, map(appByMajor, origin))
+}
+
+const listen = (account, workspace, authToken, {timeout: {duration, action}, origin, callback}) => {
```
firstdoit: Are there cases when `action` and `callback` are different?

src/courier.js
```diff
@@ -44,17 +47,22 @@ const stopAndLog = (log) => {
...
     ? {message} : message
   if (isLogLevelInfo && isSpinnerActive()) {
     if (level === 'error') {
       stopAndLog(consumeChangeLog())
       clearAbove()
       return console.log(`\n${text}\n`)
     }
-    setSpinnerText(text, timeout)
+
+    if (timeout) {
```
tamorim: That's ok, msgTimeout it's not meant to trigger the courier timeout.
As you see below, it's was created for controlling the spinner on `watch`

src/courier.js
```diff
@@ -63,38 +71,67 @@ const logToConsole = (level, origin, message) => {
   levelFormat[level](`[${time}] ${text}`)
 }
 
-const originMatch = (origin) => {
-  const {vendor, name, version} = manifest
-  const local = `${vendor}.${name}@${version.substring(0, 0)}`
-  return local === origin.substring(0, origin.indexOf('@') + 1)
+const appByMajor = (app) => {
```
tamorim: Fair enough! A `locator.js` file?

src/courier.js
```diff
@@ -63,38 +71,67 @@ const logToConsole = (level, origin, message) => {
...
+  const versionIndex = app.indexOf('@')
+  return versionIndex > -1
+    ? app.substring(0, versionIndex + 2) : app
 }
 
-const listen = (account, workspace, authToken) => {
-  let es = new EventSource(`${courierHost}/${account}/${workspace}/app-events?level=${log.level}`, {
+const getOriginByManifest = () => {
+  const {name, vendor, version} = manifest.name
```
tamorim: Oopsie :see_no_evil:

src/courier.js
```diff
@@ -63,38 +71,67 @@ const logToConsole = (level, origin, message) => {
...
+  return `${vendor}.${name}@${version}`
+}
+
+const originMatch = (msgOrigin, origin = []) => {
+  const msgOriginByMajor = appByMajor(msgOrigin)
+  if (origin.length === 0) {
+    origin.push(getOriginByManifest())
+  }
+  return any(app => msgOriginByMajor.indexOf(app) > -1, map(appByMajor, origin))
```
tamorim: Yup, much better haha

src/courier.js
tamorim: For the `install` and `uninstall` cases, no.
But I think that may exist other cases... The callback is for the `workspace:changed` event, specifically, while the timeout `action` is for any timeout that might happen.
I think it's possible someone wanting to handle these stuff differently.

src/modules/apps/uninstall.js
tamorim: Yeah, I felt dirty doing it :frowning_face:

src/modules/apps/uninstall.js
tamorim: It was supposed to hint `decremented` >:

src/courier.js
firstdoit: OK. Nevertheless, the feedback remains: I would not easily imagine the difference between them.

PR #250, opened 2017-02-07
Switches to new context-bound clients

rdumont:
To upgrade the `@vtex/api` package to its newest version.

Standardizing client usage.

By running every command and testing different scenarios. I have already tested many of them.

#### Types of changes
- [x] Refactoring (non-breaking change which doesn't change functionality)

### DIFF
--- a/src/clients.js
+++ b/src/clients.js
@@ -1,33 +1,34 @@
 /* @flow */
-import AppRegistryClient from '@vtex/api/lib/AppRegistryClient'
-import AppEngineClient from '@vtex/api/lib/AppEngineClient'
-import WorkspacesClient from '@vtex/api/lib/WorkspacesClient'
-import RouterClient from '@vtex/api/lib/RouterClient'
-import VTEXIDClient from '@vtex/api/lib/VTEXIDClient'
+import type {RegistryInstance, AppsInstance, WorkspacesInstance, RouterInstance} from '@vtex/api'
+import {Registry, Apps, Workspaces, Router, ID} from '@vtex/api'
 import endpoint from './endpoint'
 import {version} from '../package.json'
+import {getAccount, getWorkspace} from './conf'
 
 const options = {
+  account: getAccount(),
+  workspace: getWorkspace(),
   authToken: 'abc123',
   userAgent: `Toolbelt/${version}`,
+  region: 'aws-us-east-1',
 }
 
-export function registry (): AppRegistryClient {
-  return new AppRegistryClient(endpoint('registry'), options)
+export function registry (): RegistryInstance {
+  return Registry({...options, endpoint: endpoint('registry')})
 }
 
-export function apps (): AppEngineClient {
-  return new AppEngineClient(endpoint('apps'), options)
+export function apps (): AppsInstance {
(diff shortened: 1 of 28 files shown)


src/endpoint.js
```diff
@@ -1,19 +1,19 @@
 const env = process.env.VTEX_ENV === 'beta' ? 'BETA' : 'STABLE'
 
-export default function endpoint (api = 'api') {
+export default function (api = 'api') {
```
firstdoit: Named functions make for better stack traces. :)

src/clients.js
```diff
@@ -1,33 +1,34 @@
...
+  workspace: getWorkspace(),
   authToken: 'abc123',
   userAgent: `Toolbelt/${version}`,
+  region: 'aws-us-east-1',
 }
 
-export function registry (): AppRegistryClient {
-  return new AppRegistryClient(endpoint('registry'), options)
+export function registry (): RegistryInstance {
+  return Registry({...options, endpoint: endpoint('registry')})
```
firstdoit: If the options are fixed, I think we should export a simple object. This way we don't recreate clients each time.

src/modules/apps/list.js
```diff
@@ -29,14 +28,8 @@ export default {
...
-      appEngineClient().listApps(
-        getAccount(),
-        getWorkspace()
-      ).then(prop('data')),
-      appEngineClient().listLinks(
-        getAccount(),
-        getWorkspace()
-      ).then((linkedApps) => linkedApps.map(parseLocator)),
+      apps().listApps().then(prop('data')),
+      apps().listLinks().then((linkedApps) => linkedApps.map(parseLocator)),
```
firstdoit: Here we could `apps()` once and destructure the functions out of it:
`const {listApps, listLinks} = apps()`

src/modules/apps/watch.js
```diff
@@ -87,9 +85,9 @@ export default {
     return listLocalFiles(root)
     .tap(files => log.debug('Sending files:', '\n' + files.join('\n')))
     .then(mapFileObject)
-    .then(files => publishApp(files, version))
+    .then(files => registry().publishApp(files, version))
     .then(() =>
-      installApp(id)
+      apps().installApp(id)
```
firstdoit: Here, on the same chain, we end up creating two `apps`.

src/modules/infra/update.js
```diff
@@ -12,10 +11,9 @@ export default {
   handler: async function () {
     setSpinnerText('Getting available updates')
     startSpinner()
-    const [account, workspace] = [getAccount(), getWorkspace()]
     const [availableRes, installedRes] = await Promise.all([
       router().listAvailableServices(),
-      router().listInstalledServices(account, workspace),
+      router().listInstalledServices(),
```
firstdoit: Also creating duplicated objects here.

src/clients.js
firstdoit: Actually, returning a fixed object would even allow other code to simply further destructure the import, achieving what we wanted in the first place:

`import {apps: {installApp}} from '../clients'`

PR #280, opened 2017-04-07
Implemented feature to install apps from other registries, and to pub…

edufgf:
To allow users to install apps  from other registries and to publish apps to other registries.

Users can install apps from different registries and publish to different registries.

commandline

#### Types of changes

### DIFF
--- a/src/clients.ts
+++ b/src/clients.ts
@@ -21,6 +21,11 @@ const interceptor = (client) => new Proxy({}, {
   },
 })
 
+export const accountRegistry = (acc: string): Registry => {
+  if (!acc) acc = getAccount();
+  return Registry({...options, account: acc, endpoint: endpoint('registry')})
+}
+
 const [apps, router, registry, workspaces] = getToken()
   ? [
     Apps({...options, endpoint: endpoint('apps')}),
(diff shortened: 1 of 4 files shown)


src/clients.ts
```diff
@@ -21,6 +21,11 @@ const interceptor = (client) => new Proxy({}, {
   },
 })
 
+export const accountRegistry = (acc: string): Registry => {
```
amoreira: could be only `account` :)

export const accountRegistry = (account: string): Registry => {
     if (!account) account = getAccount();
   return Registry({...options, account, endpoint: endpoint('registry')})
}

src/modules/apps/install.ts
```diff
@@ -63,8 +71,7 @@ export default {
 
     const app = optionalApp || `${manifest.vendor}.${manifest.name}@${manifest.version}`
     const apps = [app, ...options._.slice(ARGS_START_INDEX)].map(arg => arg.toString())
-
     log.debug('Installing app(s)', apps)
-    return installApps(apps)
+    return installApps(apps, options.r || options.registry)
```
amoreira: I think it should exist one `default` option to registries, currently we are using 'smartcheckout' as a default registry.

```js
return installApps(apps, options.r || options.registry || 'smartcheckout')
```

PR #449, opened 2018-12-05
Fetch typings for react apps

vcalasans:
This commit affects the `vtex link` command in the following way:

- In the beginning of the linking process, checks for app dependencies in
  manifest.json. These dependencies are then added to react/package.json as
  `devDependencies`, pointing to endpoints in vtex.asset-server (https://github.com/vtex/asset-server/pull/12).
- Also asks vtex.builder-hub if it injects any dependencies in this app. vtex.builder-hub answer contains only dependencies that export components that the user is supposed to use in his app (only render-runtime, for the time being).
- `yarn --force` is run in the `react` folder, in order to fetch typings and
  install other dependencies.

The above procedures are carried out once for each `vtex link`.

#### Test it:

- Clone asset-server (vtex/asset-server#12). Link it with `vtex link`.
- Clone builder-hub (vtex/builder-hub#272). Link it with `vtex link`.
- Clone this repo and checkout this branch.
- `yarn install && yarn watch` in this branch.
- Extract these two apps: 
[testGenerateTypings.zip](https://github.com/vtex/toolbelt/files/2750090/testGenerateTypings.zip)
[testFetchTypings.zip](https://github.com/vtex/toolbelt/files/2750091/testFetchTypings.zip)
- link the `testGenerateTypings` app. builder-hub will generate its typings.
- `alias vtex=path_to_local_repo/lib/cli.js`
- link the `testFetchTypings` app. Toolbelt will fetch its typings from asset-server. If you look in `react/index.tsx`, there will be no tsc warnings. You can also access components exported by render-runtime because their typings are in the `node_modules` folder.

#### Types of changes

### DIFF
--- a/src/modules/apps/link.ts
+++ b/src/modules/apps/link.ts
@@ -136,7 +136,7 @@ const warnAndLinkFromStart = (appId: string, builder: Builder, extraData: { link
   return null
 }
 
-const watchAndSendChanges = async (appId: string, builder: Builder, extraData : {linkConfig : LinkConfig}, manifest: Manifest, context: any): Promise<any> => {
+const watchAndSendChanges = async (appId: string, builder: Builder, extraData : {linkConfig : LinkConfig}): Promise<any> => {
   const changeQueue: Change[] = []
 
   const onInitialLinkRequired = e => {
@@ -201,7 +201,6 @@ const watchAndSendChanges = async (appId: string, builder: Builder, extraData : 
     watcher
       .on('add', (file, { size }) => size > 0 ? queueChange(file) : null)
       .on('change', (file, { size }) => {
-        if (!/package\.json$/.test(file)) { getTypings(manifest, context.account, context.workspace, context.environment) }
         return size > 0
           ? queueChange(file)
           : queueChange(file, true)
@@ -338,5 +337,5 @@ export default async (options) => {
       process.exit()
     })
 
-  await watchAndSendChanges(appId, builder, extraData, manifest, context)
+  await watchAndSendChanges(appId, builder, extraData)
 }
--- a/src/modules/apps/link.ts
+++ b/src/modules/apps/link.ts
@@ -23,7 +23,7 @@ import startDebuggerTunnel from './debugger'
 import { createLinkConfig, getIgnoredPaths, getLinkedDepsDirs, getLinkedFiles, listLocalFiles } from './file'
 import legacyLink from './legacyLink'
 import lint from './lint'
-import { checkBuilderHubMessage, pathToFileObject, isLinked, resolveAppId, showBuilderHubMessage, validateAppAction } from './utils'
+import { checkBuilderHubMessage, isLinked, pathToFileObject, resolveAppId, showBuilderHubMessage, validateAppAction } from './utils'
 
 const root = process.cwd()
 const DELETE_SIGN = chalk.red('D')
@@ -92,7 +92,6 @@ const getTypings = async (manifest: Manifest, account: string, workspace: string
       ramdaReject(isNil)
     )(manifest)
 
-
   const buildersWithAppDeps =
     pipe(
       mapObjIndexed(
@@ -101,33 +100,34 @@ const getTypings = async (manifest: Manifest, account: string, workspace: string
       ramdaReject(isEmpty)
     )(buildersWithInjectedDeps as Record<string, any>)
 
-
-  const buildersToRunYarn = await pipe(
-    mapObjIndexed(
-      async (appDeps: Record<string, any>, builder: string) => {
-        const packageJsonPath = resolvePackageJsonPath(builder)
-        if (await pathExists(packageJsonPath)) {
-          const packageJson = await readJson(packageJsonPath)
-          const oldDevDeps = packageJson.devDependencies
-          const newDevDeps = await appsWithTypingsURLs(builder, account, workspace, environment, appDeps)
-          const mergedDevDeps = merge(oldDevDeps, newDevDeps)
-          if (!equals(oldDevDeps, mergedDevDeps)) {
-            await outputJson(
-              packageJsonPath,
-              { ...packageJson, ...{ 'devDependencies': mergedDevDeps } },
-              { spaces: '\t' }
-            )
-            return builder
+  const buildersToRunYarn = ramdaReject(isNil,
+    await pipe(
+      mapObjIndexed(
+        async (appDeps: Record<string, any>, builder: string) => {
+          const packageJsonPath = resolvePackageJsonPath(builder)
+          if (await pathExists(packageJsonPath)) {
+            const packageJson = await readJson(packageJsonPath)
+            const oldDevDeps = packageJson.devDependencies
+            const newDevDeps = await appsWithTypingsURLs(builder, account, workspace, environment, appDeps)
+            const mergedDevDeps = merge(oldDevDeps, newDevDeps)
+            if (!equals(oldDevDeps, mergedDevDeps)) {
+              await outputJson(
+                packageJsonPath,
+                { ...packageJson, ...{ 'devDependencies': mergedDevDeps } },
+                { spaces: '\t' }
+              )
+              return builder
+            }
           }
+          return null
         }
-        return null
-      }
-    ),
-    values,
-    Promise.all
-  )(buildersWithAppDeps as Record<string, any>)
+      ),
+      values,
+      Promise.all
+    )(buildersWithAppDeps as Record<string, any>)
+  )
 
-  map(runYarn)(ramdaReject(isNil, buildersToRunYarn))
+  map(runYarn)(buildersToRunYarn)
 }
 
 const warnAndLinkFromStart = (appId: string, builder: Builder, extraData: { linkConfig: LinkConfig } = { linkConfig: null }) => {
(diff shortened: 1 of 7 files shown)


src/modules/apps/link.ts
```diff
@@ -29,7 +31,71 @@ const UPDATE_SIGN = chalk.blue('U')
...
+  }
+  return result
+}
+
+const mergeDevDeps = (currentPackageJson: {[key: string]: any}, newDevDepsEntries: {[key: string]: string}) =>
+  merge(currentPackageJson.devDependencies, newDevDepsEntries)
+
+const runYarn = () => {
+  log.info('Running yarn in ./react/')
+  process.chdir('./react')
```
luizerasmo: Even though this will only be run in the `react` folder for now, it seems strange that this is hard-coded. Especially since this is called from `getTypings` who has a `builder` parameter that could, at some point in the future, be other than `react`.

src/modules/apps/link.ts
```diff
@@ -29,7 +31,71 @@ const UPDATE_SIGN = chalk.blue('U')
 const stabilityThreshold = process.platform === 'darwin' ? 100 : 200
 const AVAILABILITY_TIMEOUT = 1000
 const N_HOSTS = 3
+const builderHubInjectedDepsTimeout = 2000  // 2 seconds
+const yarn = join(__dirname, '../../../node_modules/yarn/bin/yarn.js install --force')
```
luizerasmo: Not sure if the way `builder-hub` calls and pass arguments to `yarn` looks a little bit better: https://github.com/vtex/builder-hub/blob/master/service/build/yarn.ts#L76

src/modules/apps/link.ts
```diff
@@ -29,7 +31,71 @@ const UPDATE_SIGN = chalk.blue('U')
...
+      const declaredDependencies = manifest.dependencies
+      const implicitDependencies = await injectedDependencies(builder, manifest.builders[builder], workspace, account, environment)
+      const appDependencies = merge(declaredDependencies, implicitDependencies)
+      if (appDependencies) {
+        log.info('Exporting app dependencies to react/package.json')
+        const reactPackageJson = JSON.parse(readFileSync(packageJsonPath, 'utf8'))
+        const newDevDepsEntries = await appsWithTypingsURLs(builder, account, workspace, environment, appDependencies)
+        reactPackageJson.devDependencies = mergeDevDeps(reactPackageJson, newDevDepsEntries)
+        writeFileSync(packageJsonPath, JSON.stringify(reactPackageJson, null, 2))
+        runYarn()
```
luizerasmo: Maybe I got it wrong, but it seems that every time the user runs `vtex link`, we are running `yarn install --force` locally, before sending the files to `builder-hub`, right? If this is the case, it is worth noting that this adds some overhead to every link request (and if the user, for some reason, does not use `yarn`, the first `vtex link` will take a long time).

src/modules/apps/link.ts
```diff
@@ -29,7 +31,71 @@ const UPDATE_SIGN = chalk.blue('U')
...
+const resolveAppId = async (appName: string, appVersion: string) => await apps.getApp(`${appName}@${appVersion}`).then(prop('id'))
+
+const injectedDependencies = async (builderName: string, builderVersion: string, workspace: string, account: string, environment: string) => {
+  const extension = (environment === 'prod') ? 'myvtex' : 'myvtexdev'  // Remove this
+  const http = axios.create({
+    baseURL: `https://${workspace}--${account}.${extension}.com`, // change this route!!!
+    timeout: builderHubInjectedDepsTimeout,
+  })
+  try {
+    const res = await http.get(`/_v/private/builder/0/getdeps/${builderName}/${builderVersion}`)
```
luizerasmo: *Idea*: Couldn't we make a single request with the app's `manifest.json`, in such a way that `builder-hub` returns all injected deps for all builders (in this case, it would continue to be only the `react` one).

src/modules/apps/link.ts
```diff
@@ -29,7 +31,71 @@ const UPDATE_SIGN = chalk.blue('U')
 const stabilityThreshold = process.platform === 'darwin' ? 100 : 200
 const AVAILABILITY_TIMEOUT = 1000
 const N_HOSTS = 3
+const builderHubInjectedDepsTimeout = 2000  // 2 seconds
+const yarn = join(__dirname, '../../../node_modules/yarn/bin/yarn.js install --force')
+
+const resolvePackageJsonPath = (builder: string) => resolvePath(process.cwd(), `${builder}/package.json`)
+
+const resolveAppId = async (appName: string, appVersion: string) => await apps.getApp(`${appName}@${appVersion}`).then(prop('id'))
```
amoreira: this could be moved to `modules/apps/utils.ts`

src/modules/apps/link.ts
```diff
@@ -29,7 +31,71 @@ const UPDATE_SIGN = chalk.blue('U')
...
+    return res.data
+  } catch (e) {
+    return {}
+  }
+}
 
+const appTypingsURL = async (account: string, workspace: string, environment: string, appName: string, appVersion: string, builder: string): Promise<string> => {
+  const extension = (environment === 'prod') ? 'myvtex' : 'myvtexdev'  // Remove this
+  const appId = await resolveAppId(appName, appVersion)
+  const typingsPath = /\+build/.test(appId) ? 'linked/v1' : 'v1'
```
amoreira: update command already checks if some app is linked or not:

https://github.com/vtex/toolbelt/blob/b1962c05812a2708443fb42ccb218a3f8d90cb96/src/modules/apps/update.ts#L30

src/modules/apps/link.ts
```diff
@@ -29,7 +31,71 @@ const UPDATE_SIGN = chalk.blue('U')
...
+  return `https://${workspace}--${account}.${extension}.com/_v/private/typings/${typingsPath}/${appId}/${builder}` // change this route!!!!!!
+}
+
+const appsWithTypingsURLs = async (builder: string, account: string, workspace: string, environment: string, appDependencies: {[key: string]: string}) => {
```
amoreira:
```js 
reduce(async (acc, appName) => {
    const appVersion = appDependencies[appName]
    const typingUrl = await appTypingsURL(account, workspace, environment, appName, appVersion, builder)
    assoc(typingUrl, 3, acc)
    return acc
  }, {}, keys(appDependencies))
```

src/modules/apps/link.ts
```diff
@@ -29,7 +31,71 @@ const UPDATE_SIGN = chalk.blue('U')
...
+
+const runYarn = () => {
+  log.info('Running yarn in ./react/')
+  process.chdir('./react')
+  execSync(yarn, {stdio: 'inherit'})
+  process.chdir('../')
+  log.info('Finished running yarn')
+}
+
+const getTypings = async (builder: string, manifest: Manifest, account: string, workspace: string, environment: string) => {
```
amoreira:
```js

import { readJson, outputJson } from 'fs-extra'

...

const getTypings = async (builder: string, {dependencies, builders}: Manifest, account: string, workspace: string, environment: string) => {
  if (!has(builder, builders)) {
    const implicitDependencies = await injectedDependencies(builder, builders[builder], workspace, account, environment)
    const appDependencies = merge(dependencies, implicitDependencies)

    if(appDependencies) {
      const packageJsonPath = resolvePackageJsonPath(builder)
      const reactPackageJson = await readJson(resolvePackageJsonPath(builder)).catch(() => ({}))

      const newDevDepsEntries = await appsWithTypingsURLs(builder, account, workspace, environment, appDependencies)
      const devDependencies = mergeDevDeps(reactPackageJson, newDevDepsEntries)

      await outputJson(packageJsonPath, {
        ...reactPackageJson,
        ...devDependencies
      })
      runYarn()
    }
  }
}
```

src/modules/apps/link.ts
```diff
@@ -29,7 +31,71 @@ const UPDATE_SIGN = chalk.blue('U')
...
+  if (existsSync(packageJsonPath)) {
+    if (has(builder, manifest.builders)) {
+      const declaredDependencies = manifest.dependencies
+      const implicitDependencies = await injectedDependencies(builder, manifest.builders[builder], workspace, account, environment)
+      const appDependencies = merge(declaredDependencies, implicitDependencies)
+      if (appDependencies) {
+        log.info('Exporting app dependencies to react/package.json')
+        const reactPackageJson = JSON.parse(readFileSync(packageJsonPath, 'utf8'))
+        const newDevDepsEntries = await appsWithTypingsURLs(builder, account, workspace, environment, appDependencies)
+        reactPackageJson.devDependencies = mergeDevDeps(reactPackageJson, newDevDepsEntries)
```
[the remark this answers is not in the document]
vcalasans: Sorry, my bad.

src/modules/apps/link.ts
vcalasans: Cool, thank you! But would this not cause problems because of the `async` function you pass as first argument to `reduce`? The reason I wrote the simple `for` loop here was because I tested it first with a functional approach and it seemed that ramda did not work very well with promises. What would you suggest?

src/modules/apps/link.ts
vcalasans: This is now changed. `yarn install --force` is only called if the content of `devDependencies` in `package.json` changes.

src/modules/apps/link.ts
vcalasans: Done

src/modules/apps/link.ts
vcalasans: Moved isLinked to src/modules/apps/utils.ts and using it in link.ts.

src/modules/apps/link.ts
vcalasans: The `execSync` function does not support an array of options for the command. That said, I made the `runYarn` function a little cleaner now.

PR #453, opened 2018-12-13
Add support for builder-hub messages

vcalasans:
This commit adds support for VTEX/builder-hub messages, replacing
the `toggle` routes.

Check https://github.com/vtex/builder-hub/pull/280 for more information.

#### Types of changes

### DIFF
--- a/src/modules/apps/utils.ts
+++ b/src/modules/apps/utils.ts
@@ -148,7 +148,7 @@ export function optionsFormatter(billingOptions: BillingOptions) {
 
 export async function checkBuilderHubMessage(cliRoute: string): Promise<any> {
   const http = axios.create({
-    baseURL: `https://calasans--basedevmkp.myvtexdev.com`,
+    baseURL: `https://vtex.myvtex.com`,
     timeout: 10000,
   })
   try {
@@ -177,7 +177,7 @@ export async function showBuilderHubMessage(message: string, showPrompt: boolean
         throw new CommandError(`${appNameInput} doesn't match with the app name.`)
       }
     } else {
-      console.log(message)
+      log.info(message)
     }
   }
 }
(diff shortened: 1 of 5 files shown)


src/modules/apps/utils.ts
```diff
@@ -144,3 +145,39 @@ export function optionsFormatter(billingOptions: BillingOptions) {
...
+  if(message) {
+    if (showPrompt) {
+      const confirmMsg = `Are you absolutely sure?\n${message ? message : ''}\nPlease type in the name of the app to confirm (ex: vtex.getting-started):`
+      const appNameInput = await promptConfirm(confirmMsg)
+      const AppName = `${manifest.vendor}.${manifest.name}`
+      if (appNameInput !== AppName) {
+        throw new CommandError(`${appNameInput} doesn't match with the app name.`)
+      }
+    } else {
+      console.log(message)
```
luizerasmo: Rather than `console.log`, we should use the `log` (from `./src/logger`), right? (and then we may decide if we use `log.warn`, `log.info`, etc.

src/modules/apps/utils.ts
```diff
@@ -144,3 +145,39 @@ export function optionsFormatter(billingOptions: BillingOptions) {
   table.push([{ content: chalk.bold('Terms of use:'), hAlign: 'center' }, { content: billingOptions.termsURL, hAlign: 'center' }])
   return table.toString()
 }
+
+export async function checkBuilderHubMessage(cliRoute: string): Promise<any> {
+  const http = axios.create({
+    baseURL: `https://vtex.myvtex.com`,
+    timeout: 10000,
```
luizerasmo: How about changing this timeout depending on the route (`link` versus `publish`)?

PR #457, opened 2019-01-03
Adds support app initialization to toolbelt

pedro823:
To add the possibility to `vtex init` a [support app](https://github.com/vtex-apps/hello-support).

Soon enough, users will be able to configure support apps. IO infrastructure is working on getting that set up. Toolbelt should be ready to create those apps.
[Read more about support apps at the support app README](https://github.com/vtex-apps/hello-support/blob/master/README.md) 

- Link (in the yarn sense) this version of toolbelt: `cd <toolbelt-folder> && yarn link`.
Run `vtex init` and try to link (in the VTEX IO sense) the support app.

#### Types of changes


src/modules/init/index.ts
```diff
@@ -15,10 +15,7 @@ const currentFolderName = process.cwd().replace(/.*\//, '')
 const templates = {
   'react getting-started': 'render-getting-started',
   'graphql getting-started': 'product-review-graphql-example',
-  'react+graphql': 'catalogue',
-  'hello graphql': 'hello-graphql',
-  'hello react': 'hello-react',
-  'dreamstore getting-started': 'dreamstore-getting-started',
+  'react+graphql': 'render-guide',
```
igorframos: VTEX is very against merge commits, so you should probably rebase from master instead of adding a merge commit.

pedro823: I accidentally made the change in the browser. I'll rebase instead

PR #458, opened 2019-01-14
Handle error when publish fails

vcalasans:
This commit changes toolbelt's behaviour whenever a `publish` command fails:

- Stop throwing 'Something exploded'.
- Reduce the amount of error messages that appear.

#### Types of changes

### DIFF
--- a/src/modules/apps/publish.ts
+++ b/src/modules/apps/publish.ts
@@ -122,10 +122,7 @@ const publisher = (workspace: string = 'master') => {
             spinner.succeed(`${appId} was published successfully!`)
         }
       } catch (e) {
-        spinner.fail(`Fail to publish ${appId}`)
-        log.error(e.message)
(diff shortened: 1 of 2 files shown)


src/modules/apps/publish.ts
```diff
@@ -122,10 +122,7 @@ const publisher = (workspace: string = 'master') => {
             spinner.succeed(`${appId} was published successfully!`)
         }
       } catch (e) {
-        spinner.fail(`Fail to publish ${appId}`)
-        log.error(e.message)
```
luizerasmo: Are we sure we want to remove this `log.error(e.message)`?

Example: the 'minimum toolbelt version' error message - would the user be able to see that he/she had an obsolete `toolbelt` version when `publish`ing an app?

luizerasmo: Ah, nevermind... I saw the `builder-hub` PR now...

PR #459, opened 2019-01-17
Implement `vtex undeprecate`

vcalasans:
Used to revert deprecation of apps. The usage is exactly the same as the `vtex deprecate` command.

The code is also almost the same as the 'deprecate.ts' module, except that we need to manually implement a http request, because `node-vtex-api` does not expose any method for undeprecating apps.

In the near future we will implement an `undeprecateApp` method in `node-vtex-api`, upgrade the version of `node-vtex-api` used in this project and use it instead of the `axios` http request.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vtex",
-  "version": "2.35.0",
+  "version": "2.36.0",
   "description": "The platform for e-commerce apps",
   "main": "lib/index.js",
   "bin": "lib/cli.js",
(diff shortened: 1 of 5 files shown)


victorges:
What do you think of something like `vtex deprecate --undo` or some flag like that instead?

vcalasans:
> What do you think of something like vtex deprecate --undo or some flag like that instead?

According to the team's discussion, we are leaving this as `vtex undeprecate` for the time being.

PR #461, opened 2019-01-17
Close chrome tab on mac after successful login

thiagomurakami:
Add script to close Chrome tab that was successful on login. Only works on mac.

Tab that is left open on mac is annoying.

?

#### Types of changes

### DIFF
--- a/scripts/closeChrome.scpt
+++ b/scripts/closeChrome.scpt
@@ -6,7 +6,7 @@ end is_running
 on run argv
 	-- Check if chrome is running
 	set chromeRunning to is_running("Google Chrome")
-	-- Set URL to the once gived in arguments
+	-- Set URL to the one given in arguments
 	set lookupUrl to item 1 of argv
 
 	-- Engine
--- a/scripts/closeChrome.scpt
+++ b/scripts/closeChrome.scpt
@@ -1,0 +1,32 @@
+-- Check if app is running
+on is_running(appName)
+	tell application "System Events" to (name of processes) contains appName
+end is_running
+
+on run argv
+	-- Check if chrome is running
+	set chromeRunning to is_running("Google Chrome")
+	-- Set URL to the once gived in arguments
+	set lookupUrl to item 1 of argv
+
+	-- Engine
+	if chromeRunning then
+		tell application "Google Chrome"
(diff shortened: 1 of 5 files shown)


scripts/closeChrome.scpt
```diff
@@ -0,0 +1,32 @@
...
+						tell w to activate
+
+						set (active tab index of window j) to i
+						return
+					end if
+				end repeat
+			end repeat
+		end tell
+	end if
+end run
```
pedro823: is... this... appleScript??
Will this run only on macs? What about windows/linux users?

src/modules/auth/login.ts
```diff
@@ -92,9 +107,10 @@ export default async (options) => {
   const account = optionAccount || (usePrevious && cachedAccount) || await promptAccount(cachedAccount && optionWorkspace)
   const workspace = optionWorkspace || (usePrevious && cachedWorkspace) || 'master'
   try {
-    const { login, token } = await authAndSave(account, workspace, optionWorkspace)
+    const { login, token, returnUrl } = await authAndSave(account, workspace, optionWorkspace)
     log.debug('Login successful', login, account, token, workspace)
     log.info(`Logged into ${chalk.blue(account)} as ${chalk.green(login)} at workspace ${chalk.green(workspace)}`)
+    closeChromeTabIfMac(returnUrl)
```
pedro823: would really appreciate other OSes too.

scripts/closeChrome.scpt
thiagomurakami: Yes :O
I couldn't do this by javascript on `auth-server` 😢 , so the alternative was opening a PR here...
I don't even have any windows or linux environment configured to develop for those operating systems :(

Do you use linux? Want to help me on this one? :)

pedro823: I'll try doing this later in bash. Don't know if it'll work, though.

pedro823: Oh. Turns out `pkill --newest chrome` should do, IF the person didn't open any new tabs while logging in.

firstdoit:
Or @pedro823 can you test whether pkill works on linux and mac (and firefox)

PR #462, opened 2019-01-23
Validate expiration of token in checkLogin

victordamatta:
Do a simple check on the expiration of local token when using the cli.

Saves people some time since the local token will always be valid (if it is not valid, it will immediately prompt for login).

Run 'vtex local token' with an expired token configured.

#### Types of changes

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -51,13 +51,12 @@ const logToolbeltVersion = () => {
   log.debug(`Toolbelt version: ${pkg.version}`)
 }
 
-const validToken = (): boolean => {
+const hasValidToken = (): boolean => {
   const token = getToken()
   if (!token) { return false }
 
   const decoded = decode(token)
-  if (decoded === null || typeof decoded === 'string' || decoded.exp === undefined) { return false }
-  if (Number(decoded.exp) < (Date.now() / 1000)) { return false }
+  if (!decoded || !decoded.exp || Number(decoded.exp) < (Date.now() / 1000)) { return false }
 
   return true
 }
@@ -65,7 +64,7 @@ const validToken = (): boolean => {
 const checkLogin = args => {
   const first = args[0]
   const whitelist = [undefined, 'login', 'logout', 'switch', 'whoami', 'init', '-v', '--version']
-  if (!validToken() && whitelist.indexOf(first) === -1) {
+  if (!hasValidToken() && whitelist.indexOf(first) === -1) {
     log.debug('Requesting login before command:', args.join(' '))
     return run({ command: loginCmd })
   }
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -51,12 +51,12 @@ const logToolbeltVersion = () => {
   log.debug(`Toolbelt version: ${pkg.version}`)
 }
 
-const validToken = () => {
+const validToken = (): boolean => {
   const token = getToken()
   if (!token) { return false }
 
-  const decoded = decode(token, {complete: true})
-  if (Number(decoded.payload.exp) < (Date.now() / 1000)) { return false }
+  const decoded = decode(token)
+  if (decoded === null || typeof decoded === 'string' || decoded.exp === undefined || Number(decoded.exp) < (Date.now() / 1000)) { return false }
 
   return true
 }
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -56,7 +56,8 @@ const validToken = (): boolean => {
   if (!token) { return false }
 
   const decoded = decode(token)
-  if (decoded === null || typeof decoded === 'string' || decoded.exp === undefined || Number(decoded.exp) < (Date.now() / 1000)) { return false }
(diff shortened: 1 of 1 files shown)


src/cli.ts
```diff
@@ -50,10 +51,20 @@ const logToolbeltVersion = () => {
   log.debug(`Toolbelt version: ${pkg.version}`)
 }
 
+const validToken = () => {
+  const token = getToken()
+  if (!token) { return false }
+
+  const decoded = decode(token, {complete: true})
```
pedro823: Can you be sure the token always returns decoded? what if `getToken()` returns an empty string, or a malformed token?

src/cli.ts
```diff
@@ -50,10 +51,20 @@ const logToolbeltVersion = () => {
   log.debug(`Toolbelt version: ${pkg.version}`)
 }
 
+const validToken = () => {
```
pedro823:
```ts
const validToken = (): boolean => {
```

victordamatta:
[ch3860]

src/cli.ts
pedro823: (because in the next line you're trying to access `decoded.payload.exp`)

jsonwebtoken.decode has this signature:
```ts
decode(token: string, options?: DecodeOptions): string | { [key: string]: any; }
```

src/cli.ts
```diff
@@ -50,10 +51,21 @@ const logToolbeltVersion = () => {
   log.debug(`Toolbelt version: ${pkg.version}`)
 }
 
+const validToken = (): boolean => {
+  const token = getToken()
+  if (!token) { return false }
+
+  const decoded = decode(token)
+  if (decoded === null || typeof decoded === 'string' || decoded.exp === undefined) { return false }
```
pedro823: no need to check for `decoded.exp`, `undefined < somethingDefined` is always `false`.
`!decoded` is somewhat more accurate than `decoded === null`, but :thisisfine:

src/cli.ts
```diff
@@ -50,10 +51,21 @@ const logToolbeltVersion = () => {
...
+  if (Number(decoded.exp) < (Date.now() / 1000)) { return false }
```
victorges: Is this using the same timezone? `Date.now()` probably returns the time in the current timezone while the timestamp in the token is probably UTC. Could you make sure there's no mismatch here?

src/cli.ts
victorges: I think the pattern in JS is to use `!value` or `value` to check if `=== null` or `!=== null`.

So **nit**: `if (!decoded || !decoded.exp) return false;` (if the object has an `exp` property, it's not a `string` either)

src/cli.ts
```diff
@@ -50,10 +51,21 @@ const logToolbeltVersion = () => {
   log.debug(`Toolbelt version: ${pkg.version}`)
 }
 
+const validToken = (): boolean => {
```
victorges: **nit**: `tokenIsValid` or `hasValidToken`?

src/cli.ts
```diff
@@ -4,6 +4,7 @@ import * as Bluebird from 'bluebird'
 import chalk from 'chalk'
 import { all as clearCachedModules } from 'clear-module'
 import { CommandNotFoundError, find, MissingRequiredArgsError, run as unboundRun } from 'findhelp'
+import { decode } from 'jsonwebtoken'
```
victorges: Is this a new dependency? Do we need to put it in `package.json`?

src/cli.ts
ArturPimentel: I think there is no need for manual checking at all. The route `https://vtexid.vtex.com.br/api/vtexid/pub/authenticated/user?authToken={{authToken}}` does all the validation we need, returning `null` on invalid tokens. This is also good because we rely on VTEX ID instead of our own design.

victorges: Downside of that is doing an asynchronous request before every toolbelt command, effectively reducing the responsiveness of the CLI. Or is this function called only for `vtex local token` command?

victordamatta: This is called before pretty much every CLI command

victordamatta: I think it would be best not to do the remote call, the only validation I care about here is if I can read the expiration date

src/cli.ts
victordamatta: Dependency was already there

src/cli.ts
victordamatta: True that

src/cli.ts
ArturPimentel: Good point. If this is meant to be executed before every toolbelt command, there is no point in checking with VTEX ID, all APIs called will do that anyway.

src/cli.ts
victordamatta: According to the JWT spec and the Date.Now() doc, both should be calculated using the Unix epoch, which is UTC

PR #468, opened 2019-02-01
Adds `support` command

pedro823:
To add the possibility of calling `vtex support <account>`, asking `vtex.support-authority` for a new token with the desired account and a support role. This will only work if there are support apps of __your vendor__ installed in the master workspace of the supported account.

This will enable toolbelt to use support apps in production. For more information about support apps, [click here (VTEX devs only)](https://docs.google.com/document/d/19HsGMa9YI_ykMt-0Y1QLFO2zJBpGy52ezCQLPqqvdIg/edit#heading=h.zfxvki4r5d89).

It's easier if you ask me to try it with you! But...

- Make sure you're using **STAGING**.
- Login into account 1, create workspace.
- Clone & `vtex link` `vtex.support-authority`.
- Clone & `vtex link` `vtex.hello-support`.
- `vtex infra install 0.17.1-dev`.
- Login into account 2, create workspace, repeat steps 3-5.
- From account 1, do `vtex support account-2`.
- choose `Support Agent`.
- You'll be logged in in account 2, as a support agent.
- do `vtex ls`! Mess around. You're not allowed to `vtex link` though (Your permissions are defined in `hello-support`).
- When you're done, `vtex switch account-1` again :)

https://asciinema.org/a/KKmTgdqezVByTHNf5MoWFyoDD

#### Types of changes
- [x] Requires change to documentation, which **Hasn't been updated while support apps aren't live yet.**


src/modules/support/login.ts
```diff
@@ -0,0 +1,92 @@
...
+import { getAccount, getToken, getWorkspace, saveAccount, saveToken, saveWorkspace } from '../../conf'
+import * as env from '../../env'
+import log from '../../logger'
+
+const getAvailableRoles = async (region: string, token: string, supportedAccount: string): Promise<string[]> => {
+  const response = await axios.get(
+    `http://support-authority.vtex.${region}.vtex.io/${getAccount()}/${getWorkspace()}/${supportedAccount}/roles`,
+    {
+      headers: {
+        'Authorization': token,
```
terciodemelo: Should this have an explicit authorization type?

src/cli.ts
```diff
@@ -64,9 +64,7 @@ const hasValidToken = (): boolean => {
   if (!token) { return false }
 
   const decoded = decode(token)
-  if (!decoded || !decoded.exp || Number(decoded.exp) < (Date.now() / 1000)) { return false }
-
-  return true
+  return decoded && typeof decoded !== 'string' && decoded.exp && Number(decoded.exp) >= Date.now() / 1000
```
terciodemelo: This is kinda strange, I don't get what `decoded.exp` is supposed to be, but please double check this expression's correctness

src/modules/support/login.ts
igorframos: No need. We only have 1 type and has to be the default as many people do not add an explicit type already.

src/cli.ts
igorframos: I just inverted the condition above to return the result directly. It should be correct, or toolbelt would keep asking me to log in again, even when my token is valid.

Also, for the record: `decoded.exp` is a unix timestamp. It is directly comparable to `Date.now() / 1000`, which is a number as well.
JS console showing that both values are comparable

PR #470, opened 2019-02-06
Remove builder hub from legacy link if

firstdoit:
@marcoskwkm 🙏


firstdoit:
HISTORY BEING MADE!!!

PR #471, opened 2019-02-08
Enable linking in subdirectories

lucasecdb:
Add possibility to link in any subdirectory of an app.

It was kind of annoying to go to the root of the app to be able to link.

 - Generate a executable running `yarn build`
 - Inside any app, go to a subdirectory and run `~/path/to/toolbelt/lib/cli.js link`

#### Types of changes

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -1,3 +1,4 @@
+import { accessSync } from 'fs'
 import { readFile } from 'fs-extra'
 import * as path from 'path'
 import { memoize } from 'ramda'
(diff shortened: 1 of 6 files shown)


src/manifest.ts
```diff
@@ -1,26 +1,44 @@
...
 
 const readFileUtf = async (file: string): Promise<string> => {
-  try {
-    return await readFile(file, 'utf8')
-  } catch (e) {
-    throw new CommandError(`Manifest file doesn't exist or is not readable. Please add a manifest.json file in the root of the app folder.`)
+  return await readFile(file, 'utf8')
+}
+
+export const manifestFileName = 'manifest.json'
```
lbebber: ⛏ `MANIFEST_FILE_NAME` perhaps

src/manifest.ts
```diff
@@ -1,26 +1,44 @@
...
+
+export const getAppRoot = () => {
+  const cwd = process.cwd()
+
+  const find = dir => {
+    try {
+      accessSync(path.join(dir, manifestFileName))
+      return dir
+    } catch (e) {
+      if (dir === '/') {
```
lbebber: does this work on windows? (i.e. is the root directory `/`?)

src/manifest.ts
```diff
@@ -1,26 +1,44 @@
...
+        throw new CommandError(`Manifest file doesn't exist or is not readable. Please add a manifest.json file in the root of the app folder.`)
```
lbebber: This error message could/should consider the possibility the user isn't running the command on the project folder at all.

klzns:
What do you think of limiting how many levels down you could be in?

I think it just makes sense to allow linking one level deep. Example:

```
app *works*
├── graphql *works*
└── react *works*
    └── foo/ *don't work*
```

src/manifest.ts
klzns: You can use https://nodejs.org/api/path.html#path_path_sep

src/manifest.ts
lucasecdb: I've kept in camel case because some other constants in this file are also defined with that casing, but I can change this one

src/manifest.ts
lucasecdb: @klzns I don't think that works, because on windows the root would be something like `C:\\`, I could use [this](https://nodejs.org/api/path.html#path_path_parse_path) to get the root dir name, what do you think?

klzns: Hey, check my comment on the pull request.

If we set just a specific number of levels that we go up, that wouldn't be an issue.

lucasecdb:
@klzns I'd like the flexibility of linking without limitations on the level count.

Suppose you're setting up some directory structure and you need to be in the following directory

```
app/
└── react/
    └── components/
        └── MyComponent/ *here*
            └── index.js
            └── styles.css
```

I think it would be nice to be able to link inside the `MyComponent` folder, without having to go to the `react` folder

PR #473, opened 2019-02-13
Adds command `vtex browse admin`

pedro823:
To add a new set of commands: `vtex browse`. For now, it has `vtex browse admin`, but can be expanded.

Currently, it is a hassle to open admin pages when you're already on toolbelt. You need to craft the url yourself if it's your first time opening it, or search through your browser history to find the correct link. 
Now, if you're already on toolbelt, just by using `vtex browse admin`, the admin panel will open in your default browser.

- download this branch and `yarn watch` it.
- login to any account/workspace, in prod or staging. It should work everywhere.
- Use `vtex browse admin`, watch magic happen.

#### Types of changes

### DIFF
--- a/src/modules/browse/admin.ts
+++ b/src/modules/browse/admin.ts
@@ -1,24 +1,15 @@
 import * as opn from 'opn'
 import * as conf from '../../conf'
 
-const buildPrefix = (account: string, workspace: string): string => {
-  const base = 'https://'
-  const workspacePrefix = workspace === 'master' ? '' : `${workspace}--`
-  return base + workspacePrefix + account
-}
-
-const buildDomain = (region: conf.Environment): string =>
+const chooseDomain = (region: conf.Environment): string =>
   region === conf.Environment.Production
-    ? '.myvtex.com'
-    : '.myvtexdev.com'
+    ? 'myvtex.com'
+    : 'myvtexdev.com'
 
 export default () => {
   const region = conf.getEnvironment()
   const { account, workspace } = conf.currentContext
-  const prefix = buildPrefix(account, workspace)
-  const domain = buildDomain(region)
-  const path = '/admin'
-  const uri = prefix + domain + path
+  const uri = `https://${workspace}--${account}.${chooseDomain(region)}/admin`
 
   opn(uri, { wait: false })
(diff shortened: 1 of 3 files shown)


src/modules/browse/admin.ts
```diff
@@ -0,0 +1,24 @@
+import * as opn from 'opn'
+import * as conf from '../../conf'
+
+const buildPrefix = (account: string, workspace: string): string => {
+  const base = 'https://'
+  const workspacePrefix = workspace === 'master' ? '' : `${workspace}--`
```
igorframos: Since you are sending the user to their current workspace, you should always set the prefix. Keep in mind that no prefix is not the same as master, it means any workspace that is set to receive production traffic and has positive weight. If you want to use the current workspace set in the context, you should make `master` explicit as well.

From my understanding, this will probably make this function a simple one-liner that you can put directly in the place that is using it. 🙂

src/modules/browse/admin.ts
```diff
@@ -0,0 +1,24 @@
...
+  return base + workspacePrefix + account
+}
+
+const buildDomain = (region: conf.Environment): string =>
```
igorframos: This is more a `chooseDomain` than a `buildDomain`, I think. 🤔

src/modules/browse/admin.ts
```diff
@@ -0,0 +1,24 @@
...
+    ? '.myvtex.com'
+    : '.myvtexdev.com'
+
+export default () => {
+  const region = conf.getEnvironment()
+  const { account, workspace } = conf.currentContext
+  const prefix = buildPrefix(account, workspace)
+  const domain = buildDomain(region)
+  const path = '/admin'
+  const uri = prefix + domain + path
```
igorframos: You can probably make everything a one-liner and not lose much readability here.
```
const uri = `${workspace}--${account}.${domain(region)}/admin`
```
Just be careful not to duplicate any separators like the dot between account and domain. 🙂

src/modules/browse/admin.ts
pedro823: Well, that's neat. Fixed

PR #475, opened 2019-02-14
Fix init for store-theme

firstdoit:


src/modules/init/index.ts
```diff
@@ -82,13 +93,13 @@ const promptContinue = async (repoName: string) => {
   }
 }
 
-const manifestFromPrompt = async () => {
-  return mapSeries([
+const manifestFromPrompt = async (repo: string) => {
+  return mapSeries<any, string>([
```
ArturPimentel: What does this mean? That we're mapping functions that receive anything and return a string?

firstdoit: I.......


don't know

🤣

PR #485, opened 2019-02-21
Improve unlink error message when app is not found in the workspace

vcalasans:
Adds a better error message whenever the user tries to unlink an app that is not linked in the current workspace.

#### Types of changes
- [x] Minor change


src/modules/apps/unlink.ts
```diff
@@ -17,7 +17,7 @@ const unlinkApp = async (app: string) => {
     log.info('Successfully unlinked', app)
   } catch (e) {
     if (e.response.status === 404) {
-      log.error(`${app} was not found in the current workspace. \
+      log.error(`${app} is not linked in the current workspace. \
```
klzns: 👌

PR #488, opened 2019-02-25
Implement --no-watch option for `vtex link`

vcalasans:
As the title says, this PR implements an option for the `vtex link` command.

Issuing `vtex link --no-watch` will make the command exit as soon as the app has finished the linking process.

#### Types of changes

### DIFF
--- a/src/modules/apps/link.ts
+++ b/src/modules/apps/link.ts
@@ -345,9 +345,8 @@ export default async (options) => {
     if (options.watch === false) {
       await listenBuild(subject, buildTrigger, { waitCompletion: true })
       return
-    } else {
-      unlistenBuild = await listenBuild(subject, buildTrigger, { waitCompletion: false, onBuild, onError }).then(prop('unlisten'))
     }
+    unlistenBuild = await listenBuild(subject, buildTrigger, { waitCompletion: false, onBuild, onError }).then(prop('unlisten'))
   } catch (e) {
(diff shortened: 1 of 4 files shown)


src/modules/apps/link.ts
```diff
@@ -342,8 +342,12 @@ export default async (options) => {
   try {
     const buildTrigger = performInitialLink.bind(this, appId, builder, extraData)
     const [subject] = appId.split('@')
-    const { unlisten } = await listenBuild(subject, buildTrigger, { waitCompletion: false, onBuild, onError })
-    unlistenBuild = unlisten
+    if (options.watch === false) {
```
pedro823: `if (!options.watch)` doesn't work? BTW, can't you deconstruct this option in the beggining of the function?

src/modules/apps/link.ts
```diff
@@ -342,8 +342,12 @@ export default async (options) => {
...
+      await listenBuild(subject, buildTrigger, { waitCompletion: true })
+      return
+    } else {
```
pedro823: `else` after return

src/modules/apps/link.ts
lbebber: ☝️ `if (!options.watch)` will evaluate to `true` if `watch` is nil. I think a strict comparison with `false` might be good in this case?

PR #495, opened 2019-02-27
Fix post release

vcalasans:
This PR fixes the following `vtex release` issue:

- Post release scripts output were not being correctly redirected to stdio. 

#### Types of changes

### DIFF
--- a/src/modules/release/utils.ts
+++ b/src/modules/release/utils.ts
@@ -68,7 +68,8 @@ const runCommand = (
 ) => {
   let output
   try {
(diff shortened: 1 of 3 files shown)


src/modules/release/utils.ts
```diff
@@ -68,7 +68,8 @@ const runCommand = (
 ) => {
   let output
   try {
-    output = execSync(cmd, {stdio: hideOutput ? 'pipe' : ['inherit', 'pipe']})
+    console.log(`running ${cmd}`)
```
lbebber: is this console.log left here intentionally?

vcalasans: Already removed. Tks

PR #510, opened 2019-03-18
Add VTEX IO cluster targeting

terciodemelo:
This PR intends to enable VTEX Toolbelt to target any VTEX IO cluster
by providing the routing cookie on some routes, and adding the
cluster id domain infix in other routes.

To allow VTEX Toolbelt to communicate with any VTEX IO clutser
This generalizes the communication between toolbelt and vtex io clusters, in such a way that we're no longe limited to two environments (prod and staging), but we can out-of-the-box to use it against any created environment

Basically execute common operations against multiple clusters, by specify the environment variable `VTEXIO_REGION`

#### Types of changes

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -27,7 +27,7 @@ export function publicEndpoint(): string {
 }
 
 export function clusterIdDomainInfix(): string {
-  return process.env.VTEX_REGION ? `.${region()}` : ''
+  return process.env.VTEX_REGION ? `.${process.env.VTEX_REGION}` : ''
 }
 
 export function envCookies(): string {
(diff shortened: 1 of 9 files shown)


src/env.ts
```diff
@@ -25,3 +25,11 @@ export function region(): string {
 export function publicEndpoint(): string {
   return getEnvironment() === Environment.Staging ? 'myvtexdev.com' : 'myvtex.com'
 }
+
+export function clusterIdDomainInfix(): string {
```
igorframos: What do you think of simply changing the `publicEndpoint()` function to return everything, including this region infix?

terciodemelo: I thought about it, but most of the time we don't want the infix, because the cookie header will do it and axios lib won't handle the `set-cookie` by cloudfront. But when using `opn` you go to the browser and will get the cookie by going through the redirection process.

Now that you said it, I double checked and I should remove the infix from `sse.ts` and use the cookie header in `createEventSource`

PR #513, opened 2019-03-18
Improve logger

vcalasans:
This PR makes some changes to toolbelt's logging facilities:

- Loggers are now built using the `createLogger` function from `winston`, which receives _transports_ that log differently depending on the log scope.
- Every log (starting on debug level) will now be stored locally in a `vtex_debug.json` file.
- The `simpleConsoleTransport` is used for every other command, as well as for `vtex link --verbose`. It replicates the functionality of toolbelt's previous default logger.

#### Types of changes

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -1,6 +1,73 @@
-import * as winston from 'winston'
+import chalk from 'chalk'
+import { join } from 'path'
+import { createLogger, format, transports } from 'winston'
+import * as Transport from 'winston-transport'
+import { configDir } from './conf'
 
-// TODO: Configure transport to send errors to Splunk
-winston.cli()
+// Setup logging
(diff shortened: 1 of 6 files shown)


src/logger.ts
```diff
@@ -1,6 +1,72 @@
...
-// TODO: Configure transport to send errors to Splunk
-winston.cli()
+// Setup logging
+const VERBOSE = '--verbose'
+const isVerbose = process.argv.indexOf(VERBOSE) >= 0
 
-export default winston
+const { combine, timestamp, colorize } = format
+
+export const DEBUG_LOG_FILE_PATH = join(configDir, 'vtex_debug.txt')
```
iagoaraujo: *nit*: can you put a comment above this line with the complete path of this file?

PR #519, opened 2019-03-25
Replace inquirer with enquirer

vcalasans:
This PR replaces the `inquirer` dependency with `enquirer` as an attempt to solve the freeze in output reported by some Windows users.

#### Types of changes
- [x] Refactor


src/modules/init/index.ts
```diff
@@ -88,7 +92,7 @@ const promptTemplates = async (): Promise<string> => {
 }
 
 const promptContinue = async (repoName: string) => {
-  const proceed = prop('proceed', await inquirer.prompt({
+  const proceed = prop('proceed', await enquirer.prompt({
```
lbebber: why are you using `enquirer.prompt` here instead of using `promptConfirm`?

lbebber: same applies to other similar lines

vcalasans: because these are not confirmation prompts!

vcalasans: Sry, this was is indeed a `confirm` prompt! Will fix it.

PR #540, opened 2019-04-24
Persist StickyHost between links and publishes

victorcolombo:
Locally store a host used in a link or publish. If another link or publish is done within a certain threshold, try to reuse that host.

By reusing a host, we can take of advantage of caches from previous builds.

``vtex link``
``vtex publish``

#### Types of changes

### DIFF
--- a/src/conf.ts
+++ b/src/conf.ts
@@ -27,6 +27,9 @@ export const saveWorkspace = (workspace = 'master') =>
 export const saveEnvironment = (env: Environment) =>
   conf.set('env', env)
 
+export const saveStickyHost = (appId: string, stickyHost: string) =>
+  conf.set(`${appId}.sticky-host`, {stickyHost, lastUpdated: Date.now()})
+
 export const getAll = (): any => conf.all
 
 export const getAccount = (): string =>
@@ -40,6 +43,12 @@ export const getToken = (): string =>
 
 export const getWorkspace = (): string =>
   conf.get('workspace')
+
+export const getStickyHost = (appId: string): {stickyHost: string; lastUpdated: Date} =>
+  conf.get(`${appId}.sticky-host`)
+
+export const hasStickyHost = (appId: string): boolean =>
+  conf.has(`${appId}.sticky-host`)
 
 const envFromProcessEnv = {
   'beta': Environment.Staging,
--- a/src/conf.ts
+++ b/src/conf.ts
@@ -27,8 +27,8 @@ export const saveWorkspace = (workspace = 'master') =>
 export const saveEnvironment = (env: Environment) =>
   conf.set('env', env)
 
-export const saveStickyHost = (appId: string, stickyHost: string) =>
-  conf.set(`${appId}.sticky-host`, {stickyHost, lastUpdated: Date.now()})
+export const saveStickyHost = (appName: string, stickyHost: string) =>
+  conf.set(`apps.${appName}.sticky-host`, {stickyHost, lastUpdated: Date.now()})
 
 export const getAll = (): any => conf.all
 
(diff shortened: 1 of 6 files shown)


src/host.ts
```diff
@@ -85,3 +89,23 @@ export const getMostAvailableHost = async (
...
+export const getSavedOrMostAvaliableHost = async (
+  appId: string,
+  builder: Builder,
+  nHosts: number,
+  timeout: number
+): Promise<string> => {
+  if (hasStickyHost(appId)) {
+    log.debug(`Found sticky host saved locally`)
+    const {stickyHost, lastUpdated} = getStickyHost(appId)
+    const timeElapsed = moment.duration(moment(new Date()).diff(lastUpdated))
```
vcalasans: You can use `moment.now()` here

src/host.ts
```diff
@@ -2,9 +2,13 @@ import { Builder } from '@vtex/api'
 import * as retry from 'async-retry'
 import { map, reduce } from 'ramda'
 
+import * as moment from 'moment'
+import { getStickyHost, hasStickyHost, saveStickyHost } from './conf'
 import { BuilderHubTimeoutError } from './errors'
 import log from './logger'
 
+const TTL_SAVED_HOST_HOURS = 6
```
vcalasans: Shouldn't this be 4?

src/host.ts
```diff
@@ -85,3 +89,23 @@ export const getMostAvailableHost = async (
 
   return stickyHint
 }
+
+export const getSavedOrMostAvaliableHost = async (
```
vcalasans: `*Avaliable*` -> `*Available*`

src/host.ts
victorcolombo: I'm not sure which number to pick here. Is there any reason why 4 would be better/more suitable than 6?

vcalasans: I just thought we agreed on 4 before. But we can leave it at 6.

src/host.ts
```diff
@@ -85,3 +89,24 @@ export const getMostAvailableHost = async (
...
+    const {stickyHost, lastUpdated} = getStickyHost(appId)
+    const timeElapsed = moment.duration(moment().diff(lastUpdated))
+    if (timeElapsed.asHours() <= TTL_SAVED_HOST_HOURS) {
+      return stickyHost
+    }
+    log.debug(`Saved sticky host has expired`)
+  }
+  log.debug(`Finding a new sticky host`)
+  const newStickyHost = await getMostAvailableHost(appId, builder, nHosts, timeout)
+  saveStickyHost(appId, newStickyHost)
```
vcalasans: You should check if the `newStickyHost` is `undefined` before saving it. There are cases where the `availability` requests might fail (timeout).

src/conf.ts
```diff
@@ -27,6 +27,9 @@ export const saveWorkspace = (workspace = 'master') =>
 export const saveEnvironment = (env: Environment) =>
   conf.set('env', env)
 
+export const saveStickyHost = (appId: string, stickyHost: string) =>
+  conf.set(`${appId}.sticky-host`, {stickyHost, lastUpdated: Date.now()})
```
vcalasans: I was checking the `vtex.json` file where `configstore` saves the configs. I think configstore is messing up with the `.`'s in your strings and interpreting them as "attributes of an object". The relevant part in my `vtex.json` looks like:
```
		"render-ssr@2": {
			"1": {
				"1": {
					"sticky-host": {
						"stickyHost": "ExFdGkgBTFlfBVdLGVFMUkwHHwoNG1MZCFheSEdCWRQGBFRPAFFbAgBYUAoMXV8FUVEBHlcBB1ZcAV5TAgFB",
						"lastUpdated": 1556212228504
					}
				}
			}
		}
	}

```
Do you think we can make this cleaner? Also, can you remove the app's version before saving the stickyHost, so we can keep the stickyHost across different versions of the same app? I know there is a function somewhere in toolbelt's code that does that.

PR #544, opened 2019-04-30
vtex whoami shows better message if unable to read workspace state

vcalasans:
If toolbelt is unable to read the current workspace's state, it must be logged out of the platform. This PR makes the `vtex whoami` message replace the `Logged into` message part by `Not logged in. Previously logged into` in this case.

#### Types of changes

### DIFF
--- a/src/greeting.ts
+++ b/src/greeting.ts
@@ -28,7 +28,7 @@ export const greeting = async (): Promise<string[]> => {
     let loggedMessage = 'Logged into'
     let state = await getWorkspaceState()
     if (!state) {
-      loggedMessage = 'Not logged in. Previously logged into'
+      loggedMessage = `${chalk.red('Not logged in')}. Previously logged into`
       state = ''
     }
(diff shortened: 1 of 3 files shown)


src/greeting.ts
```diff
@@ -17,16 +17,21 @@ const getWorkspaceState = async (): Promise<string> => {
...
 }
 
 export const greeting = async (): Promise<string[]> => {
   if (account && login && workspace) {
-    const state = await getWorkspaceState()
-    return [`Logged into ${chalk.blue(account)} as ${chalk.green(login)} at ${chalk.yellowBright(state)}workspace ${chalk.green(workspace)} in environment ${chalk.red(environment)}`]
+    let loggedMessage = 'Logged into'
+    let state = await getWorkspaceState()
+    if (!state) {
+      loggedMessage = 'Not logged in. Previously logged into'
```
lbebber: Maybe turn `Not logged in` red?

vcalasans: good idea!

PR #551, opened 2019-05-08
Always run yarn locally on link and publish (react and node folders)

vcalasans:
As the titles says, to avoid inconsistencies between `yarn.lock` and `package.json` when the code is processed by `builder-hub`.

#### Types of changes

### DIFF
--- a/src/modules/utils.ts
+++ b/src/modules/utils.ts
@@ -84,7 +84,6 @@ export const formatNano = (nanoseconds: number): string =>
     0
   )}ms`
 
-
 export const runYarn = (relativePath: string, force: boolean) => {
   log.info(`Running yarn in ${chalk.green(relativePath)}`)
   const root = getAppRoot()
--- a/src/modules/utils.ts
+++ b/src/modules/utils.ts
@@ -1,6 +1,11 @@
+import chalk from 'chalk'
+import { execSync } from 'child-process-es6-promise'
+import { existsSync } from 'fs'
+import { resolve as resolvePath } from 'path'
 import { currentContext } from '../conf'
 import { BuildFailError } from '../errors'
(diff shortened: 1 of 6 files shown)


src/modules/utils.ts
```diff
@@ -76,3 +83,29 @@ export const formatNano = (nanoseconds: number): string =>
...
+  execSync(
+    command,
+    {stdio: 'inherit', cwd: resolvePath(root, `${relativePath}`)}
+  )
+  log.info('Finished running yarn')
+}
+
+export const runYarnIfPathExists = (relativePath: string) => {
+  const root = getAppRoot()
+  const pathName = resolvePath(root, `${relativePath}/`)
```
marcoskwkm: ditto

src/modules/utils.ts
```diff
@@ -76,3 +83,29 @@ export const formatNano = (nanoseconds: number): string =>
...
+
+export const runYarn = (relativePath: string, force: boolean) => {
+  log.info(`Running yarn in ${chalk.green(relativePath)}`)
+  const root = getAppRoot()
+  const command = force ?
+    `${yarnPath} --force --non-interactive` :
+    `${yarnPath} --non-interactive`
+  execSync(
+    command,
+    {stdio: 'inherit', cwd: resolvePath(root, `${relativePath}`)}
```
marcoskwkm: Doesn't simply `resolvePath(root, relativePath)` work here?

src/modules/utils.ts
vcalasans: thx!

src/modules/utils.ts
vcalasans: thx!

PR #563, opened 2019-05-28
Sponsor Account

vcalasans:
Implements new commands for sponsor account functionality:

- `vtex edition`: gets the edition of the current account.
- `vtex edition set {edition@version}`: sets the edition of the current account. 

### How to test: 

Use account `vtexgame42` and edition `vtexgame43.edition-test@0.x`.

#### Types of changes

### DIFF
--- a/src/clients/abTester.ts
+++ b/src/clients/abTester.ts
@@ -1,4 +1,4 @@
-import { forWorkspace, IODataSource } from '@vtex/api'
+import { AppClient, InstanceOptions, IOContext } from '@vtex/api'
 
 const routes = {
   Abort: (workspace: string) => `${routes.ABTester}/finish/${workspace}`,
@@ -8,9 +8,10 @@ const routes = {
   Status: () => `${routes.ABTester}/status`,
 }
 
-export class ABTester extends IODataSource {
-  protected httpClientFactory = forWorkspace
-  protected service = 'ab-tester.vtex'
+export class ABTester extends AppClient {
+  constructor(context: IOContext, options: InstanceOptions) {
+    super('ab-tester.vtex', context, options)
+  }
 
   // Abort AB Test in a workspace.
   public finish = async (workspace: string) =>
--- a/src/clients/abTester.ts
+++ b/src/clients/abTester.ts
@@ -10,7 +10,7 @@ const routes = {
 
 export class ABTester extends AppClient {
   constructor(context: IOContext, options: InstanceOptions) {
-    super('ab-tester.vtex', context, options)
+    super('vtex.ab-tester', context, options)
   }
 
   // Abort AB Test in a workspace.
--- a/src/clients/sponsor.ts
+++ b/src/clients/sponsor.ts
@@ -27,14 +27,14 @@ export class Sponsor extends IOClient {
     this.http.get(this.routes.getEdition, { metric: 'get-edition' })
 
   public setEdition = async (account: string, edition: string) =>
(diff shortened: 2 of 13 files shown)


src/clients/abTester.ts
```diff
@@ -8,9 +8,10 @@ const routes = {
   Status: () => `${routes.ABTester}/status`,
 }
 
-export class ABTester extends IODataSource {
-  protected httpClientFactory = forWorkspace
-  protected service = 'ab-tester.vtex'
+export class ABTester extends AppClient {
+  constructor(context: IOContext, options: InstanceOptions) {
+    super('ab-tester.vtex', context, options)
```
victorges: Is this correct? Maybe the `AppClient` needs to receive the `vendor.name` in the correct order `(vtex.ab-tester`)

src/clients/sponsor.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  public setEdition = async (account: string, edition: string) =>
+    this.http.put(this.routes.setEdition(account), { edition }, { metric: 'set-edition' })
+
+  public runHouseKeeper = async () =>
+    this.http.post(this.routes.runHouseKeeper, {}, { metric: 'run-house-keeper' })
+
+  private get routes() {
+    return {
+      getSponsorAccount: `http://kube-router.${this.region}.vtex.io/_account/${this.account}`,
```
victorges: I think `apps`, `kube-router` and `housekeeper` deserved separate clients as they are actually separate services

src/modules/sponsor/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+
+
+const DEFAULT_TIMEOUT = 10000
+
+export const options = {
+  timeout: (envTimeout || DEFAULT_TIMEOUT) as number,
+  retries: 3,
+}
+
+export const getIOContext = () => ({
```
victorges: Wasn't this declared anywhere else before? 🤔

src/modules/sponsor/setEdition.ts
```diff
@@ -0,0 +1,38 @@
...
+  )
+  if (!proceed) {
+    throw new UserCancelledError()
+  }
+  await switchAccount(sponsorAccount, {})
+}
+
+export default async (edition: string) => {
+  const previousConf = conf.getAll()
+  const previousAccount = conf.getAccount()
```
victorges: **nit**: I guess you can use `previousConf.account`

src/clients/abTester.ts
vcalasans: You are right! Forgot to dis-invert this.

src/clients/sponsor.ts
vcalasans: Considering our case here, I thought it would be good to have routes specific to the sponsor account functionality bundled in the same client. But it is true that this makes the code less maintainable. I will add these functions to the clients in `@vtex/api` in a later PR.

src/modules/sponsor/utils.ts
vcalasans: I don't think so, not exactly like this. But it would be good to put this in a more global `utils.ts` and use it elsewhere in the code. It would clean up some other parts of the code. Will do it in later PRs.

src/modules/sponsor/setEdition.ts
vcalasans: Sure

PR #571, opened 2019-06-05
Improve `vtex update` command

vcalasans:
This PR uses the new housekeeper API to resolve which updates are to be made to the workspace. It then shows tables containing the updates to be made, separated by the app installation type (infra, user-installed, edition-installed), and prompts if the user wants to proceed with the update.

image

#### Types of changes


src/modules/apps/update.ts
```diff
@@ -1,67 +1,90 @@
...
+const sourceFilter = (source: string) => filter((obj: { source: string }) => includes(prop('source', obj), [source]))
 
-const extractAppLocator = pipe(prop('app'), parseLocator)
+const includes = (k: string, list: string[]) => list.indexOf(k) >= 0
 
-const updateVersion = (app) => {
-  app.version = app.latest
-  return app
-}
+const printAppsDiff = (resolvedUpdates: any, message: string, path: string[], filterFunction?: (obj: any) => boolean, pluckFunction?: (obj: any) => any) => {
```
iagoaraujo: I think this method can be more simple.
It can receive only the `resolvedUpdates`, `message` and `type` (this one can be "apps" or  "infra"). Them this own method defines which is the filterFunction considering the value of `type`.

src/modules/apps/update.ts
```diff
@@ -1,67 +1,90 @@
...
+  return true
+}
 
-  console.log(`${table.toString()}\n`)
+export default async () => {
+  const housekeeper = new Housekeeper(getIOContext(), IOClientOptions)
+  const getSpinner = ora('Getting available updates').start()
+  const resolvedUpdates = await housekeeper.resolve()
+  getSpinner.stop()
+  const hasAvailableUpdates =
```
iagoaraujo: The names of `print` methods are suggesting they only have to print something, but actually they also are returning a boolean when there are available updates.
What do you think about separate these two things: print and check available updates?  `hasAvailableUpdates` can be only a check on each field resolvedUpdates to check if the length isn't zero.
 If there are availableUpdates you can can a method `printResolvedUpdates`. So this one will run the four `printAppsDiff`.

PR #574, opened 2019-06-12
Enforce pinned dependencies

victorcolombo:
Rewrite pinned dependency versions on `package.json`

Follow up to https://github.com/vtex/builder-hub/pull/589

#### Types of changes

### DIFF
--- a/src/modules/apps/link.ts
+++ b/src/modules/apps/link.ts
@@ -229,7 +229,7 @@ export default async (options) => {
     const pinnedDeps : Map<string, string> = new Map(Object.entries(aux))
     await bluebird.map(buildersToRunLocalYarn, fixPinnedDependencies(pinnedDeps), { concurrency: 1 })
   } catch(e) {
-    log.info('Failed to check for pinned dependencies')
+    log.info('Failed to check for pinned dependencies')
   }
   // Always run yarn locally for some builders
   map(runYarnIfPathExists, buildersToRunLocalYarn)
--- a/src/modules/apps/link.ts
+++ b/src/modules/apps/link.ts
@@ -1,5 +1,7 @@
 import { Builder, Change } from '@vtex/api'
 import * as retry from 'async-retry'
+import axios from 'axios'
+import * as bluebird from 'bluebird'
 import chalk from 'chalk'
 import * as chokidar from 'chokidar'
 import * as debounce from 'debounce'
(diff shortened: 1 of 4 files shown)


src/modules/apps/link.ts
```diff
@@ -212,6 +224,13 @@ export default async (options) => {
   if (options.setup || options.s) {
     await setup()
   }
+  try {
+    const aux = await getPinnedDependencies(builderHttp)
+    const pinnedDeps : Map<string, string> = new Map(Object.entries(aux))
+    await bluebird.map(buildersToRunLocalYarn, fixPinnedDependencies(pinnedDeps), { concurrency: 1 })
+  } catch(e) {
+    log.info('Failed to check for pinned dependencies')
```
vcalasans: Is there a strange character here in the `Failed`?

src/modules/utils.ts
```diff
@@ -112,6 +115,39 @@ export const runYarnIfPathExists = (relativePath: string) => {
...
+  }
+}
+
+export const fixPinnedDependencies = R.curry(async (pinnedDeps: Map<string, string>, relativePath: string) => {
+  const jsonPath = resolvePath(getAppRoot(), `${relativePath}/package.json`)
+  if (!await pathExists(jsonPath)) {
+    return
+  }
+  const packageJSON = JSON.parse((await readFile(jsonPath)).toString())
+  const dependencies = new Map(Object.entries(packageJSON.dependencies))
```
vcalasans: This throws an error if `packageJSON.dependencies` is `undefined`.

src/modules/utils.ts
```diff
@@ -112,6 +115,39 @@ export const runYarnIfPathExists = (relativePath: string) => {
...
+  const devDependencies = new Map(Object.entries(packageJSON.devDependencies))
```
vcalasans: This throws an error if `packageJSON.devDependencies` is `undefined`

src/modules/apps/link.ts
victorcolombo: Ops

src/modules/utils.ts
```diff
@@ -112,6 +115,53 @@ export const runYarnIfPathExists = (relativePath: string) => {
...
+  if (!await pathExists(jsonPath)) {
+    return
+  }
+  const packageJSON = JSON.parse((await readFile(jsonPath)).toString())
+  const dependencies = getEntryMapFromObject(packageJSON, 'dependencies')
+  const devDependencies = getEntryMapFromObject(packageJSON, 'devDependencies')
+  const outdatedDeps = leftMapDifference(dependencies, pinnedDeps)
+  const outdatedDevDeps = leftMapDifference(devDependencies, pinnedDeps)
+  const newPackageJSON = R.reduce((obj, dep) => {
+    log.warn(`${dep} is outdated. Upgrading to ${pinnedDeps.get(dep)}`)
```
vcalasans: I think we should change this warning because the user's dependencies might not necessarily be outdated.

src/modules/apps/link.ts
```diff
@@ -212,6 +224,13 @@ export default async (options) => {
   if (options.setup || options.s) {
     await setup()
   }
+  try {
+    const aux = await getPinnedDependencies(builderHttp)
+    const pinnedDeps : Map<string, string> = new Map(Object.entries(aux))
+    await bluebird.map(buildersToRunLocalYarn, fixPinnedDependencies(pinnedDeps), { concurrency: 1 })
+  } catch(e) {
+    log.info('Failed to check for pinned dependencies')
```
vcalasans: I think this message should be changed to something like `Failed setting pinned dependencies`

PR #576, opened 2019-06-14
Initialize A/B test with new API

tergol1:
Initialize A/B test adapted to the new API, which doesn't need anymore to tell the confidence level.

Now `ab-tester` declares a new route to initialize a test which doesn't require a value o confidence level. This PR aims to initialize the test with this new format but also maintain support for old version of `ab-tester`.

Try to initialize a A/B test with IO CLI and verify if works well.

#### Types of changes

### DIFF
--- a/src/modules/workspace/abtest/start.ts
+++ b/src/modules/workspace/abtest/start.ts
@@ -60,22 +60,28 @@ export default async () => {
   const abTesterManifest = await installedABTester()
   const workspace = await promptProductionWorkspace('Choose production workspace to start A/B test:')
 
-  if (semver.satisfies(abTesterManifest.version, '>=0.10.0')) {
-    log.info(`Setting workspace ${chalk.green(workspace)} to A/B test`)
-    await abtester.start(workspace)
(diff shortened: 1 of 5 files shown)


src/modules/workspace/abtest/start.ts
```diff
@@ -56,14 +57,23 @@ ${chalk.red(significanceLevel)} significance level. Proceed?`,
 
 
 export default async () => {
-  await checkIfABTesterIsInstalled()
+  const abTesterManifest = await installedABTester()
   const workspace = await promptProductionWorkspace('Choose production workspace to start A/B test:')
+
+  if (semver.satisfies(abTesterManifest.version, '>=0.10.0')) {
+    log.info(`Setting workspace ${chalk.green(workspace)} to A/B test`)
+    await abtester.start(workspace)
```
iagoaraujo: *nit*: as a simple client, `abtester` doesn't treat errors, just throw them. If the abtest app has specific errors which can be treated by toolbelt with better error messages, remember to take of them when you use the `start`, `startLgeacy` and `finish` methods.

PR #584, opened 2019-06-27
Allow to pass "-" to switch to last used workspace/account

kaisermann:

Allow to use `vtex switch -` and `vtex use -` to switch account/workspace back to previous one used. Just like `git checkout -` or `cd -`. 

Not exactly a problem, but it would be a nice to have.

```shell
vtex switch account1
vtex switch account2
vtex switch - # should switch to `account1`
vtex switch - # should switch to `account2`
vtex switch - # should switch to `account1`
```

```shell
vtex use workspace1
vtex use workspace2
vtex use - # should switch to `workspace1`
vtex use - # should switch to `workspace2`
vtex use - # should switch to `workspace1`
```

#### Types of changes

### DIFF
--- a/src/modules/auth/switch.ts
+++ b/src/modules/auth/switch.ts
@@ -27,8 +27,11 @@ const hasAccountSwitched = (account: string) => {
 }
 
 export default async (account: string, options) => {
-  if(account === '-') {
-    account = getLastUsedAccount() || account
+  if (account === '-') {
+    account = getLastUsedAccount()
+    if (account == null) {
+      throw new CommandError('No last used account was found')
+    }
   }
 
   const previousAccount = getAccount()
--- a/src/modules/auth/switch.ts
+++ b/src/modules/auth/switch.ts
@@ -1,7 +1,7 @@
 import chalk from 'chalk'
 
 import { split } from 'ramda'
-import { getAccount } from '../../conf'
+import { getAccount, getLastUsedAccount } from '../../conf'
 import { CommandError } from '../../errors'
 import log from '../../logger'
 import loginCmd from './login'
@@ -27,6 +27,10 @@ const hasAccountSwitched = (account: string) => {
 }
 
 export default async (account: string, options) => {
+  if(account === '-') {
+    account = getLastUsedAccount() || account
+  }
+
   const previousAccount = getAccount()
   // Enable users to type `vtex switch {account}/{workspace}` and switch
   // directly to a workspace without typing the `-w` option.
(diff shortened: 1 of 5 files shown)


kaisermann:
I'm not sure if `conf` is the best place to persist what were the last used account and workspace 🤔

lbebber:
how about saving it on a `/tmp` file?

src/modules/workspace/use.ts
```diff
@@ -26,6 +26,11 @@ export default async (name: string, options?) => {
   let production = options ? (options.p || options.production) : null
   let confirm
   const accountName = getAccount()
+
+  if(name === '-') {
```
klzns:
```suggestion
  if (name === '-') {
```

src/modules/workspace/use.ts
```diff
@@ -26,6 +26,11 @@ export default async (name: string, options?) => {
...
+    name = getLastUsedWorkspace() || name
```
klzns: I think it should throw if there's no `getLastUsedWorkspace`

src/modules/auth/switch.ts
```diff
@@ -27,6 +27,10 @@ const hasAccountSwitched = (account: string) => {
 }
 
 export default async (account: string, options) => {
+  if(account === '-') {
+    account = getLastUsedAccount() || account
```
klzns: here too

kaisermann:
> how about saving it on a `/tmp` file?

was actually thinking about that, will do!

src/modules/workspace/use.ts
athoscouto: At least `.zsh` uses the current dir as last one when there is no history.

klzns: I think it should throw because this will happen:

image

kaisermann: I was using `git`s behaviour as a reference, which uses the actual `-` as the fallback. Anyway, totally agree about the `throw`

lbebber: I don't really like git's behaviour, if that's actually the case 😅
an error would be better IMO

src/modules/auth/switch.ts
```diff
@@ -27,6 +27,13 @@ const hasAccountSwitched = (account: string) => {
 }
 
 export default async (account: string, options) => {
+  if (account === '-') {
+    account = getLastUsedAccount()
+    if (account == null) {
```
vcalasans: Isn't `if (!account)` better here?

src/modules/workspace/use.ts
```diff
@@ -26,6 +26,14 @@ export default async (name: string, options?) => {
   let production = options ? (options.p || options.production) : null
   let confirm
   const accountName = getAccount()
+
+  if (name === '-') {
+    name = getLastUsedWorkspace()
+    if (name == null) {
```
vcalasans: Isn't `if (!name)` better here?

PR #590, opened 2019-07-18
Always convert formatted log to string before calling the replace method

vcalasans:
This PR stringifies the `message` or `code` received from the logger before calling the `replace` method. 

#### Types of changes

### DIFF
--- a/src/sse.ts
+++ b/src/sse.ts
@@ -96,7 +96,7 @@ export const logAll = (context: Context, logLevel: string, id: string, senders?:
       return // Ignore logs without message or code.
     }
     const suffix = sender.startsWith(id) ? '' : ' ' + chalk.gray(sender)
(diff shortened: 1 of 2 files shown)


src/sse.ts
```diff
@@ -96,7 +96,7 @@ export const logAll = (context: Context, logLevel: string, id: string, senders?:
       return // Ignore logs without message or code.
     }
     const suffix = sender.startsWith(id) ? '' : ' ' + chalk.gray(sender)
-    const formatted = (message || code || '').replace(/\n\s*$/, '') + suffix
+    const formatted = String(message || code || '').replace(/\n\s*$/, '') + suffix
```
iagoaraujo: Why do you need Stringify it? I thought `(message || code || '')` would be always a string, otherwise you couldn't use the replace method.

PR #605, opened 2019-08-13
Import/Export redirects to vtex.rewriter

vcalasans:
This PR implements redirects import/export/delete functionality. 

The redirects are fed as `.csv` files according to the following JsonSchema:

```
{
  type: 'array',
  items: {
    type: 'object',
    properties: {
      from: {
        type: 'string',
      },
      to: {
        type: 'string',
      },
      endDate: {
        type: 'string',
      },
      type: {
        type: 'string',
        enum: ['PERMANENT', 'TEMPORARY'],
      },
    },
  }
```

#### Types of changes


src/clients/rewriter.ts
```diff
@@ -0,0 +1,34 @@
...
+  importRoutes: `/importroutes`,
+  exportRoutes: '/exportroutes',
+}
+
+export class Rewriter extends AppClient {
+  constructor(context: IOContext, options: InstanceOptions) {
+    super('vtex.rewriter', context, options)
+  }
+
+  // Abort AB Test in a workspace.
```
tergol1: Are you sure this comment is correct?

PR #606, opened 2019-08-14
Refactor setup module and add tests

tiagonapoli:
Add tests to setup module. In order to simplify testing some decoupling and refactors were made


src/modules/setup/BuilderHubDatasource.ts
```diff
@@ -0,0 +1,70 @@
...
+const SECOND = 1000
+const builderHubTimeout = 2 * SECOND
+
+const retryOpts = {
+  retries: 2,
+  minTimeout: 1000,
+  factor: 2,
+}
+
+export class BuilderHubDatasource {
```
iagoaraujo: I think this should be only `BuilderHub` and you could move it to `src/clients`.

src/modules/setup/index.ts
```diff
@@ -0,0 +1,13 @@
...
+import { setupTSConfig } from './setupTSConfig'
+import { setupTypings } from './setupTypings'
+
+const buildersToAddAdditionalPackages = ['react', 'node']
+
+export default async () => {
+  const manifest = await getManifest()
+  await setupESLint(manifest, buildersToAddAdditionalPackages)
+  await setupTSConfig(manifest)
+  await setupTypings(manifest)
```
iagoaraujo: Can you do these three steps asynchronously? I think the problem is setting the package.json, which is performed by two of them. But maybe we can have only two process running asynchronously: one updating the package and another setting up the ts config.

PR #617, opened 2019-08-28
Refactor setup module and add tests 

tiagonapoli:


src/clients/Builder.ts
```diff
@@ -0,0 +1,21 @@
...
+const SECOND = 1000
+const builderHubTimeout = 2 * SECOND
+
+const builderBaseRoute = `/_v/builder/0`
+const routes = {
+  tsConfig: `${builderBaseRoute}/tsconfig`,
+  typings: `${builderBaseRoute}/typings`,
+}
+
+export default class Builder extends BuilderBase {
```
iagoaraujo: You don't need creating a new client. Update the node-vtex-api project with the new routes and methods to use them.

src/clients/index.ts
```diff
@@ -1,9 +1,10 @@
-import { Apps, Builder, Events, InstanceOptions, IOContext, Logger, Registry, Router, Workspaces } from '@vtex/api'
+import { Apps, Events, InstanceOptions, IOContext, Logger, Registry, Router, Workspaces } from '@vtex/api'
 import { getAccount, getToken, getWorkspace } from '../conf'
 import * as env from '../env'
 import envTimeout from '../timeout'
 import userAgent from '../user-agent'
 import Billing from './billingClient'
+import Builder from './Builder'
```
iagoaraujo: You can delete this line and import the Builder from `@vtex/api` when you move the new client code to the lib.

PR #631, opened 2019-09-20
Add `vtex test` to README

thalytafabrine:
#### Types of changes

### DIFF
--- a/src/modules/tree.ts
+++ b/src/modules/tree.ts
@@ -294,7 +294,7 @@ export default {
     requiredArgs: 'account',
   },
   test: {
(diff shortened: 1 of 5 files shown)


src/modules/tree.ts
```diff
@@ -294,7 +294,7 @@ export default {
     requiredArgs: 'account',
   },
   test: {
-    description: 'Run unit tests',
+    description: 'Run your VTEX app tests',
```
iagoaraujo: *nit*: I think you can keep *app unit tests* once there isn't a specific format to our tests

PR #633, opened 2019-09-27
Ensure node version and log command and OS detail

tiagonapoli:
Now that node >= v10 is required some user errors reports are reaching us because the error is not clear at all. We though just adding `"engines": { "node": ">= 10" }` was sufficient, and in some cases are, but for some reason npm doesn't ensure node version for every install. Because of this a check was added and if node version is older than the defined on `package.json` `engines.node` a error log is showed.

Additionally, logs on the command run and OS basic information was added.
  
Install this toolbelt branch:
```
git clone git@github.com:vtex/toolbelt.git && cd toolbelt && \
git checkout feature/check-node-version && \
```

Run any `vtex` command with `node` < v10. A log `Incompatible with node < v10. Please upgrade node to major 10 or higher.` should appear and _immediately close_ the cli. If `node` > v10 it should not.

Run any command with `--verbose`. The first lines printed should print your node version, OS information and the command and args you asked to execute

#### Types of changes
- [X] Chore

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -23,7 +23,10 @@ import { isVerbose, VERBOSE } from './utils'
 
 const nodeVersion = process.version.replace('v', '')
 if (!semver.satisfies(nodeVersion, pkg.engines.node)) {
-  console.error(chalk.bold('Incompatible with node < v10. Please upgrade node to major 10 or higher.'))
+  const minMajor = pkg.engines.node.replace('>=', '')
+  console.error(
+    chalk.bold(`Incompatible with node < v${minMajor}. Please upgrade node to major ${minMajor} or higher.`)
+  )
   process.exit(1)
 }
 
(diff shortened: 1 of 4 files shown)


src/cli.ts
```diff
@@ -17,8 +17,16 @@ import { CommandError, SSEConnectionError, UserCancelledError } from './errors'
 import log from './logger'
 import tree from './modules/tree'
 import notify from './update'
+import * as semver from 'semver'
+import * as os from 'os'
 import { isVerbose, VERBOSE } from './utils'
 
+const nodeVersion = process.version.replace('v', '')
+if (!semver.satisfies(nodeVersion, pkg.engines.node)) {
```
rafaelbac: Once we are getting the required version from the `pkg.engines.node` I would suggest updating the message dynamically depending on the required minimum version. Otherwise, if we change the expected version from 10 to 12 and forget to update the message, the user may get confused if he already has version 10 installed.

What do you think?

tiagonapoli: Oops, you're right

PR #634, opened 2019-09-30
Ignore .git folder by default

lbebber:
Prevents watching .git folder

Avoids this:
Screen Shot 2019-09-30 at 16 38 39

#### Types of changes

### DIFF
--- a/src/modules/apps/file.ts
+++ b/src/modules/apps/file.ts
@@ -19,6 +19,7 @@ const defaultIgnored = [
   'package.json',
   'node_modules/**',
   '**/node_modules/**',
+  '.git/**',
 ]
 
 const services = ['react', 'render', 'masterdata', 'service']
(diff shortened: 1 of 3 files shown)


src/modules/apps/file.ts
```diff
@@ -19,6 +19,7 @@ const defaultIgnored = [
   'package.json',
   'node_modules/**',
   '**/node_modules/**',
+  '.git',
```
lbebber:
```suggestion
  '.git/',
```

src/modules/apps/file.ts
```diff
@@ -19,6 +19,7 @@ const defaultIgnored = [
   'package.json',
   'node_modules/**',
   '**/node_modules/**',
+  '.git/',
```
lbebber:
```suggestion
  '.git/**',
```

PR #643, opened 2019-10-14
Update ESLint setup to match current recipe instructions

lucasecdb:
Updates the `vtex setup` command to match the current [recipe](https://docs.google.com/document/d/1ePi5Mn3pe1dOorKrMtC-MclBDiIDimVzf0Uj61W8lA4/edit) instructions. This creates a `package.json` at the root of the app (alongside the `manifest.json`) with all ESLint dependencies.

Since ESLint newest major, they've changed how the package imports some dependencies (such as shared configs), making them break if you load your eslint plugin other than the folder that it is installed. For example, if you have you eslint and dependencies installed in a `node` folder, and you open the repo in the root, some ESLint plugins may not work properly, because they are installed in a child directory, not the one that the editor plugin is running.

Build the toolbelt and run `vtex setup` in any IO app. It should generate a `.eslintrc.json`, `package.json` and some other files in the same folder as your `manifest.json`. If you have a react app, it should also add a `.eslintrc.json` in the `react` folder.

#### Types of changes

### DIFF
--- a/src/modules/setup/setupESLint.ts
+++ b/src/modules/setup/setupESLint.ts
@@ -4,15 +4,50 @@ import * as R from 'ramda'
 import log from '../../logger'
 import { getAppRoot } from '../../manifest'
 import { yarnPath } from '../utils'
-import { esLintrcEditor, packageJsonEditor } from './utils'
+import { esLintrcEditor, packageJsonEditor, eslintIgnoreEditor, prettierrcEditor } from './utils'
 
-const addToPackageJson = {
+const basePackageJson = (appName: string) => ({
+  name: appName,
+  private: true,
+  license: 'UNLICENSED',
+  scripts: {
+    lint: 'eslint --ext js,jsx,ts,tsx .',
+  },
+})
+
+const eslintIgnore = `
+node_modules
+`.trim()
+
+const basePrettierRc = {
+  semi: false,
+  singleQuote: true,
+  trailingComma: 'es5',
+  eslintIntegration: true,
+}
+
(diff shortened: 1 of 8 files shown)


src/modules/setup/setupESLint.ts
```diff
@@ -21,42 +54,62 @@ const addToEslintrc = {
...
   })
 }
 
-const createESLintSetup = async (lintDeps: string[], builder: string) => {
+const createESLintSetup = (appName: string, lintPackages: string[]) => {
   try {
-    const devDependencies = R.prop('devDependencies', packageJsonEditor.read(builder)) || {}
-    if (R.difference(lintDeps, R.intersection(lintDeps, R.keys(devDependencies))).length !== 0) {
-      yarnAddESLint(builder)
+    const originalRootPackageJson = packageJsonEditor.read('.')
```
tiagonapoli: I think something like this will be necessary here:
```typescript
  let originalRootPackageJson = {}
  try {
    originalRootPackageJson = packageJsonEditor.read('.')
  } catch (e) {
    if (e.code === 'ENOENT') {
      log.warn(`No package.json found in ${packageJsonEditor.path('.')}. Creating one.`)
    } else {
      log.error(e)
      return
    }
  }
```

tiagonapoli: If a `package.json` doesn't exist in root an error is being thrown right away

lucasecdb: done! :-)

PR #648, opened 2019-10-24
Fix error when running on TTY

lucasecdb:
Skip the "copy to clipboard" function on systems that don't support it.

When running the `clipboardy.writeSync` function on systems that don't have a display attached (e.g. WSL, or when running on TTY) or a X.Org server running, the function throws an error because it uses `xsel` under the hood.

Run the `vtex local token` command on a TTY. No error should be thrown and the token should be displayed as normal.

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## [2.77.13] - 2019-10-28
 ### Fixed
 - Error `Unhandled exception` when generating local token/workspace/account, due to nonexistent display.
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -5,9 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-
 ### Fixed
 - Remove CHANGELOG.md from list of ignored files
+- Error `Unhandled exception` when generating local token/workspace/account, due to nonexistent display.
```
tiagonapoli: Remember to move this line to unreleased

tiagonapoli:
Maybe you could update `clipboardy` to the newest version too? 
https://github.com/sindresorhus/clipboardy/releases
The API didn't change, and now it has typescript definitions

PR #649, opened 2019-10-24
Use `tenant-provisioner`'s `migrate` API to set account's editions

vcalasans:
This PR makes toolbelt use the specialized `tenant-provisioner` app to set account's editions.

### DIFF
--- a/src/clients/sponsor.ts
+++ b/src/clients/sponsor.ts
@@ -20,8 +20,15 @@ export class Sponsor extends IOClient {
 
   public getEdition = async () => this.http.get(this.routes.getEdition, { metric: 'get-edition' })
 
-  public setEdition = async (account: string, edition: string) =>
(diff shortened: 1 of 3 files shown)


src/clients/sponsor.ts
```diff
@@ -20,8 +20,15 @@ export class Sponsor extends IOClient {
 
   public getEdition = async () => this.http.get(this.routes.getEdition, { metric: 'get-edition' })
 
-  public setEdition = async (account: string, edition: string) =>
-    this.http.put(this.routes.setEdition(account), { edition }, { metric: 'set-edition' })
+  public setEdition = async (account: string, editionApp: string) => {
+    const [ edition, version ] = editionApp.split('@')
```
tiagonapoli: nit: The variable edition is used? If it's not it could be `_`

PR #656, opened 2019-11-14
Download own app types instead of adding it to package.json

tiagonapoli:
- Download own app types instead of adding it to `package.json`
- Change `vteximg` URL to `vtexassets`, as required from io-infra-team
- Add new options to `vtex setup`: 
  - `--types`: setup only types
  - `--own-types`: setup only the types for the app itself
- Refactor types setup to use more of an OOP approach (open to suggestions to improve the design)
- Add a `postinstall` script on react and node builders `package.json`, if they are present: The `postinstall` script will run when `yarn add`, `yarn install` or `yarn` are executed. Unfortunately I didn't find a way to run it when `yarn remove`. The script is set to `vtex setup --types --own-types`, so after the yarn command completes the types bundle for the app is downloaded and added to the `node_modules`. This is necessary because `yarn` uses the `package.json` as source of truth to the `node_modules`, and when it installs or removes something, the `node_modules` is reset.

  
The workflow for developing a Node + GraphQL app is somewhat complicated. When developing, the URL that was inserted on `package.json` regarding the app GraphQL types were a linked app URL. Before publishing and merging the GitHub branch, the developer had to remove this URL to prevent a linked app URL to go to master, but after publishing they had to run `vtex setup --ignore-linked` to add the published URL types to `package.json`. This way the app types on `package.json` would maintain consistency.

This problem would be solved if the types for itself weren't added on `package.json`. 

Install this branch:
```
cd toolbelt && \
git checkout feat/refactor-types && \
```
Run `vtex setup` in some apps, linked included, and check if the types for the app itself are on `node_modules`, but not in `package.json`.

#### Types of changes


src/modules/setup/index.ts
```diff
@@ -1,17 +1,67 @@
...
+    }
+  }
+
+  if (packageJson.scripts && packageJson.scripts.postinstall === script) {
+    return
+  }
+
+  packageJson.scripts = {
+    ...packageJson.scripts,
+    postinstall: script,
```
rafaelbac: This way we may overwrite the user's `postinstall` script. What do you think about the following?
1. Add our own script if the current one is blank
2. Ignore if the current script already contains ours
3. If there is content and our script is not there, ask the user if we can add it at the beginning of the command

PR #661, opened 2019-11-27
Notify IO when link connection is closed

iagoaraujo:
Use the colossus feature which allows us sends an event when the link connection is closed.
https://github.com/vtex/colossus/pull/287

Now we can notify the users when the "link experience" is not turned on anymore sending an event, which can be catch by an app and shows some kind of notification on browser.

- Install colossus@0.36.0-beta
- Use this toolbelt to link some app
- Close the terminal or just turn off the link
- Check on splunk: `index=colossus account=<account> workspace=<workspace> key=link_interrupted`

You can also link an app which is expected to receives a `link_interrupted` event from colossus.

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.78.3] - 2019-12-02
+
 ### Added
 - Use `onUnsubscribe` event feature to notify users about links closing connection.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Use feature of onUnsubscribe event to notify users about connection closed of links.
```
tiagonapoli: I think it would be better like this:
```
- Use `onUnsubscribe` event feature to notify users about links closing connection.
```

PR #664, opened 2019-11-28
Add a message notifying about HK being off

igorframos:
Add a message saying HK is not running on Black Friday.

People should know they need to take manual action to have their apps updated.

Publish an app.

#### Types of changes

### DIFF
--- a/src/modules/apps/publish.ts
+++ b/src/modules/apps/publish.ts
@@ -125,6 +125,7 @@ const publisher = (workspace: string = 'master') => {
           )
         } else {
           spinner.succeed(`${appId} was published successfully!`)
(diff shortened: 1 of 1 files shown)


src/modules/apps/publish.ts
```diff
@@ -125,6 +125,7 @@ const publisher = (workspace: string = 'master') => {
           )
         } else {
           spinner.succeed(`${appId} was published successfully!`)
+          log.info(`For Black Friday, we have turned off automatic update of apps. Log into the accounts you wish to update, create a workspace, install ${appId}, check it works and then promote the workspace.`)
```
klzns:
```suggestion
          log.info(`For Black Friday, we have turned off automatic update of apps. Log into the accounts you wish to update, create a production workspace, install ${appId}, check if it works and then promote the workspace.`)
```

PR #665, opened 2019-12-02
Implement feedback asking functionality

vcalasans:
This feature will make toolbelt ask for user feedback whenever he issues any command. 

If the user accepts giving feedback, his browser will automatically open this [Google Form](https://forms.gle/CRRHn6P3x9AeaWTQ8). It will then ask for feedback again in 3 months.

If the user does not want to give feedback, he will be prompted about being reminded again in the future.

#### Types of changes

### DIFF
--- a/src/nps.ts
+++ b/src/nps.ts
@@ -1,0 +1,53 @@
+import * as opn from 'opn'
+import * as enquirer from 'enquirer'
+import { prop, keys } from 'ramda'
+import * as moment from 'moment'
+
+import { getNextFeedbackDate, saveNextFeedbackDate } from './conf'
+import { promptConfirm } from './modules/prompts'
+
+const NEVER = 'NEVER'
+const NPSFormURL = 'https://docs.google.com/forms/d/1wTBsqoK11kG10-6bnuha5yzocipX_5lKG226vRNYsIk/edit?ts=5dded52b'
+
+const choices = {
+  'In 1 day': [1, 'days'],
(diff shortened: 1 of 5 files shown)


src/nps.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (shouldOpenFeedbackForm) {
+      // Ask for feedback again in 3 months.
+      saveNextFeedbackDate(
+        moment()
+          .add(3, 'months')
+          .toISOString()
+      )
+      opn(NPSFormURL, { wait: false })
+    } else {
+      let remindChoice = await enquirer
```
tiagonapoli: How about using:
```typescript
const { choice: remindChoice } = await enquirer.prompt({
          name: 'choice',
          message: 'When would you like to be reminded?',
          type: 'select',
          choices: keys(choices),
 })
```

Would that work? What do you think?

src/nps.ts
```diff
@@ -0,0 +1,66 @@
...
+        .prompt({
+          name: 'choice',
+          message: 'When would you like to be reminded?',
+          type: 'select',
+          choices: keys(choices),
```
tiagonapoli: What do you think about using `Object.keys(choices)` here? I was reflecting about using `ramda` functions instead of native ES functions and wanted to hear opinions on it. Personally I would go for native functions, which evolve with the language and are already known by the language users -- my opinion on it too is that using ES features may not create a concise code, but I think is more readable for users not deeply experienced with functional programming.

PR #668, opened 2019-12-05
Feature/candidate release

iagoaraujo:
Create new command to validate an app. When we update the builder-hub to use the new publish API the apps won't be updated automatically anymore, even if they are stable version. The published apps will be in a "probation state", which means they can be used to run AB test using their stable versions. If the app is ready to automatically deployed in every account which it's installed (using HK), so the follow command should be executed: `vtex validate <app>@<version>`.

Related PR: https://github.com/vtex/node-vtex-api/pull/282

- Link node-vtex-api with toolbelt;
- On builder-hub: move the `vtex/api` do `dependencies` on node/package.json and also link the node-vtex-api on it;
- Install some app on master and publish a new version using toolbelt and builder-hub above: `vtex publish -w <workspace>`;
- Run `vtex update` on master. The app should *not* be available to update;
- Run `vtex validate <app>@<version>`.
- Run `vtex update` again. Now the app should be updated. 

#### Types of changes

### DIFF
--- a/src/modules/apps/validate.ts
+++ b/src/modules/apps/validate.ts
@@ -53,7 +53,7 @@ const prepareValidate = async (app, originalAccount, originalWorkspace: string):
     log.info('Successfully validated', app)
   } catch (e) {
     if (e.response && e.response.status && e.response.status === 404) {
-      log.error(`Error validating ${app}. App not found`)
+      log.error(`Error validating ${app}. App not found or already validated`)
     } else if (e.message && e.response.statusText) {
       log.error(`Error validating ${app}. ${e.message}. ${e.response.statusText}`)
       return await switchToPreviousAccount(originalAccount, originalWorkspace)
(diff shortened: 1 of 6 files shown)


src/modules/apps/validate.ts
```diff
@@ -0,0 +1,79 @@
...
+
+const prepareValidate = async (app, originalAccount, originalWorkspace: string): Promise<void> => {
+  app = await validateApp(app)
+  try {
+    log.debug('Starting to validate app:', app)
+    await validateRelease(app)
+    log.info('Successfully validated', app)
+  } catch (e) {
+    if (e.response && e.response.status && e.response.status === 404) {
+      log.error(`Error validating ${app}. App not found`)
```
tiagonapoli: I think the server is returning 404 if the app is already validate, because when validating an already validated app the error message is `... App not found`

iagoaraujo: Apps is returning the same status code to apps that already are validated and not found apps 😞 
I'm going to improve the message

PR #682, opened 2019-12-23
updateNotifier: add changelog url & improvements

### LINKED ISSUE #677: Make update notification better
Currently our update notification does not provide a link to our changelog (and I **always** read it).

npm does provide a link and uses the same package as we do. So we just need to update our code and do link npm does.

## Expected Behavior
Do like npm does:
image

## Current Behavior
image

## Possible Solution
The code from npm:
https://github.com/npm/cli/blob/897537ab08261fe4323b0d079047507dba62bc11/bin/npm-cli.js#L85-L130

## Your Environment

* Version used: v2.79.2
* Environment name and version (e.g. Chrome 39, node.js 5.4): v12.8.0
* Operating System and version (desktop or mobile): MacOS
* Link to your project: n/a

AlexxNica:
Include the changelog URL and other improvements to the update-notifier.

Fixes #677 ("Make update notification better")

Change `const notifier = updateNotifier({ pkg, updateCheckInterval: 1000 * 60 * 60 * 1 })` to `const notifier = updateNotifier({ pkg, updateCheckInterval: 1 })` and run `npm run build && node ./lib/cli.js`

**Old:**
Old
**New:**
New

#### Types of changes

#### Comments
Not the best way to do it since we're adding a lot of technical debt. It'd be better to create an issue to change the module when possible.

### DIFF
--- a/src/update.ts
+++ b/src/update.ts
@@ -1,7 +1,35 @@
+import chalk from 'chalk'
 import * as updateNotifier from 'update-notifier'
 
 import * as pkg from '../package.json'
 
+const notifier = updateNotifier({ pkg, updateCheckInterval: 1000 * 60 * 60 * 1 })
+
 export default function notify() {
-  updateNotifier({ pkg, updateCheckInterval: 1000 * 60 * 60 * 1 }).notify({ isGlobal: true, isYarnGlobal: true })
+  if (notifier.update && notifier.update.latest !== pkg.version) {
+    const oldVersion = notifier.update.current
+    const latestVersion = notifier.update.latest
+    const changelog = `https://github.com/vtex/toolbelt/releases/tag/v${latestVersion}`
+    let type = notifier.update.type
+
+    switch (type) {
+      case 'major':
+        type = chalk.red(type)
+        break
+      case 'minor':
+        type = chalk.yellow(type)
+        break
+      case 'patch':
+        type = chalk.green(type)
+        break
+    }
+
+    notifier.notify({
+      isGlobal: true,
+      isYarnGlobal: true,
+      message: `New ${type} version of ${pkg.name} available! ${chalk.dim(oldVersion)} → ${chalk.green(latestVersion)}\n
+        ${chalk.yellow('Changelog:')} ${chalk.cyan(changelog)}\n
+        Run ${chalk.green(`yarn global add ${pkg.name}`)} to update!`,
+    })
+  }
 }
--- a/src/update.ts
+++ b/src/update.ts
@@ -3,13 +3,12 @@ import * as updateNotifier from 'update-notifier'
 
 import * as pkg from '../package.json'
 
-const notifier = updateNotifier({ pkg, updateCheckInterval: 1000 * 60 * 60 * 1 })
-
 export default function notify() {
+  const notifier = updateNotifier({ pkg, updateCheckInterval: 1000 * 60 * 60 * 1 })
   if (notifier.update && notifier.update.latest !== pkg.version) {
     const oldVersion = notifier.update.current
     const latestVersion = notifier.update.latest
-    const changelog = `https://github.com/vtex/toolbelt/releases/tag/v${latestVersion}`
+    const changelog = `https://github.com/vtex/toolbelt/blob/master/CHANGELOG.md`
     let type = notifier.update.type
 
     switch (type) {
@@ -27,9 +26,11 @@ export default function notify() {
     notifier.notify({
       isGlobal: true,
       isYarnGlobal: true,
-      message: `New ${type} version of ${pkg.name} available! ${chalk.dim(oldVersion)} → ${chalk.green(latestVersion)}
-        ${chalk.yellow('Changelog:')} ${chalk.cyan(changelog)}
-        Run ${chalk.green(`yarn global add ${pkg.name}`)} to update!`,
+      message: [
+        `New ${type} version of ${pkg.name} available! ${chalk.dim(oldVersion)} → ${chalk.green(latestVersion)}`,
+        `${chalk.yellow('Changelog:')} ${chalk.cyan(changelog)}`,
+        `Run ${chalk.green(`yarn global add ${pkg.name}`)} to update!`,
+      ].join('\n'),
     })
   }
 }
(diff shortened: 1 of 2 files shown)


src/update.ts
```diff
@@ -1,7 +1,35 @@
...
 import * as pkg from '../package.json'
 
+const notifier = updateNotifier({ pkg, updateCheckInterval: 1000 * 60 * 60 * 1 })
+
 export default function notify() {
-  updateNotifier({ pkg, updateCheckInterval: 1000 * 60 * 60 * 1 }).notify({ isGlobal: true, isYarnGlobal: true })
+  if (notifier.update && notifier.update.latest !== pkg.version) {
+    const oldVersion = notifier.update.current
+    const latestVersion = notifier.update.latest
+    const changelog = `https://github.com/vtex/toolbelt/releases/tag/v${latestVersion}`
```
tiagonapoli: Unfortunately for now the release section on github doesn't contain details on the latest version :( - (we are working on a CI to automate this and improve our docs and releases)

I think for now just a url to the `master` chagelog would suffice, what do you think?

src/update.ts
```diff
@@ -1,7 +1,35 @@
+import chalk from 'chalk'
 import * as updateNotifier from 'update-notifier'
 
 import * as pkg from '../package.json'
 
+const notifier = updateNotifier({ pkg, updateCheckInterval: 1000 * 60 * 60 * 1 })
```
tiagonapoli: How about adding instantiating the notifier inside the `notify` function?

AlexxNica: Oh, my bad! Will do that asap!

src/update.ts
```diff
@@ -1,7 +1,35 @@
...
+        type = chalk.green(type)
+        break
+    }
+
+    notifier.notify({
+      isGlobal: true,
+      isYarnGlobal: true,
+      message: `New ${type} version of ${pkg.name} available! ${chalk.dim(oldVersion)} → ${chalk.green(latestVersion)}
+        ${chalk.yellow('Changelog:')} ${chalk.cyan(changelog)}
+        Run ${chalk.green(`yarn global add ${pkg.name}`)} to update!`,
```
tiagonapoli: I think the manual line breaks here are messing a little with the table format:

Screenshot from 2019-12-26 14-56-11

If the message uses `/n`, like this:

```New ${type} version of ${pkg.name} available! ${chalk.dim(oldVersion)} → ${chalk.green(latestVersion)}\n${chalk.yellow('Changelog:')} ${chalk.cyan(changelog)}\nRun ${chalk.green(`yarn global add ${pkg.name}`)} to update!```

The message will be centralized:

Screenshot from 2019-12-26 14-56-31

**Comment by tiagonapoli** [comment]
Oops, just missing changelog, you can add to the Unreleased section:
```
## [Unreleased]
### Changed
- Add URL to changelog on update notification
``
(something like this)

tiagonapoli:
Thanks so much for your contribution, we really appreciate it! Aside from the comments I did everything looks good to me. Once everything is resolved I will merge it to our repo's master and release a new version :)

src/update.ts
AlexxNica: Nice catch! That was my 1st option but ESLint screamed at me because of Prettier.

We'd have to disable prettier for this line with `// eslint-disable-next-line prettier/prettier`, if that's not a problem.
Screen Shot 2019-12-26 at 3 08 12 PM

Or we can use string concatenation instead of template literals, but I think it'd be against your style guide, wouldn't it?

src/update.ts
AlexxNica: @tiagonapoli I had to stop programming for some time and am coming back and relearning everything, and I can't remember if exporting the constant in this case would be a good practice.

What do you think? I'm already doing a quick search to see if I can find something about it.

src/update.ts
tiagonapoli: Oops, in that case try to run `yarn format` in the root of the project. This will run `prettier` to format and it will do it in a way that won't affect the string outcome.
You can too write the message by line, a way would be:
```js
notifier.notify({
      isYarnGlobal: true,
      isGlobal: true,
      message: [
        `New ${type} version of ${pkg.name} available! ${chalk.dim(oldVersion)} → ${chalk.green(latestVersion)}`,
        `${chalk.yellow('Changelog:')} ${chalk.cyan(changelog)}`,
        `Run ${chalk.green(`yarn global add ${pkg.name}`)} to update!`,
      ].join('\n'),
    })

```

AlexxNica: Prettier makes it a little strange, but you call the shots! What's best for you?
Screen Shot 2019-12-26 at 4 13 34 PM

src/update.ts
tiagonapoli: In that case you wouldn't be exporting the `notifier` constant, it will just be declared inside the `notify` function:
```js
export default function notify() {
    const notifier = updateNotifier({ pkg, updateCheckInterval: 1 })  
    if (notifier.update && notifier.update.latest !== pkg.version) {
    const oldVersion = notifier.update.current
    ...
}
``` 

This way the `updateNotifier` is only instantiated when really needed (when the notify function is called) and it will be localized inside it will really be used. Of course this comes with the price that the `updateNotifier` may be instantiated several times, if the `notify` function is called many times, but this will not happen. Each `vtex someCommand` execution will call `notify` one time.

This change however is just me nitpicking, on a second though I think it's okay to leave the notifier where it's now :)

src/update.ts
tiagonapoli: I think using the array will be prettier :)

src/update.ts
AlexxNica: Yup. We can do that!

src/update.ts
AlexxNica: I think you're right on this one, so I left it inside the function as you mentioned!

AlexxNica:
There you go! I've tried to follow your reviews, see if it's ok now.

I'm sorry for being a little slow on this, as I mentioned, I'm just now relearning everything as I had to stop coding for some time.

Thank you for your patience and guidance!

klzns:
Thanks @AlexxNica!

PR #692, opened 2020-01-02
Fix publish docs URL and warn about 'vtex validate' on publish

tiagonapoli:
If the terminal truncated the previous URL and the user clicked to open it, the URL copied to the browser could be incomplete. Using a shortener may solve the problem

```
cd toolbelt && \
git checkout fix/publish-url && \
```

Try to publish any app. The new publish process prompt will show up - then test the link.

#### Types of changes


tiagonapoli:
Since now the URL is on a new line and is shorter it's less likely to truncate, but it can still happen :(

PR #694, opened 2020-01-03
Add cluster header on clients' options

tiagonapoli:
@vtex/io-infra reported that after changing to the new URLs format changing the region via `VTEX_CLUSTER` env variable on toolbelt stopped working. That's because now to set the region it's necessary to set the `x-vtex-upstream-target` header.

Install this branch:
```
cd toolbelt && \
git checkout fix/specify-region && \
```

#### Types of changes


src/clients/index.ts
```diff
@@ -27,8 +27,13 @@ const context: IOContext = {
   platform: '',
 }
 
+const regionHeader = context.region !== Region.Production ? { 'x-vtex-upstream-target': context.region } : null
```
victorges: So can this region come from `vtex config set/get` too? 😻

tiagonapoli: I'll take a look

src/sse.ts
```diff
@@ -74,7 +74,7 @@ const onLog = (
   callback: (message: Message) => void,
   senders?: string[]
 ): Unlisten => {
-  const source = `${endpoint('colossus')}/${ctx.account}/${ctx.workspace}/logs?level=${logLevel}`
+  const source = `${colossusEndpoint()}/${ctx.account}/${ctx.workspace}/logs?level=${logLevel}`
```
victorges: This also needs to change to call the new endpoint for infra services, which in case of colossus would be `infra.io.vtex.com/colossus/v0/...`. This would also need the logic for setting the `upstream-target` header when the cluster configuration is present.

PR #696, opened 2020-01-06
Log url for auth-sse

tiagonapoli:
- Log url for auth-sse

```
cd toolbelt && \
git checkout chore/log-event-server-url && \
```
Try to `vtex login --verbose` - it will debug the auth-sse URL 

#### Types of changes
- [X] Chore

### DIFF
--- a/src/sse.ts
+++ b/src/sse.ts
@@ -1,8 +1,7 @@
 import chalk from 'chalk'
-import { compose, forEach, contains, path, pathOr } from 'ramda'
-
+import { compose, contains, forEach, path, pathOr } from 'ramda'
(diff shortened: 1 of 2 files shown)


src/sse.ts
```diff
@@ -1,8 +1,7 @@
 import chalk from 'chalk'
-import { compose, forEach, contains, path, pathOr } from 'ramda'
-
```
marcoskwkm: nit: Not sure what coding standard you're following in this repo, but usually we leave an empty line between package imports and local imports.

tiagonapoli: We'll shortly configure the linter to do this automatically

PR #715, opened 2020-01-21
Execute commands in the host container

rafaelbac:
Add a new command that allows the user to execute commands in the host container. 

By using this command you are instructing Toolbelt to connect with the host container via WebSockets so it can execute the commands sent by the user (passing the stdin or not). 

In order to debug .Net apps, we need a mechanism to execute commands in the host container.

Please see https://github.com/vtex/service-runtime-base/pull/95 for instructions on how to test this PR.

#### Types of changes

### DIFF
--- a/src/clients/runtime.ts
+++ b/src/clients/runtime.ts
@@ -46,8 +46,8 @@ export class Runtime {
       pathname: path,
       query: {
         __v: manifest.majorRange,
-        params: command?.split(' '),
-        interactive,
+        cmd: command?.split(' '),
+        stdin: interactive,
       },
     }
     const formattedUrl = url.format(urlObject)
--- a/src/clients/runtime.ts
+++ b/src/clients/runtime.ts
@@ -19,7 +19,7 @@ export class Runtime {
     this.workspace = workspace
   }
 
-  public executeCommand = async (command: string, interactive: boolean) => {
+  public async executeCommand(command: string, interactive: boolean) {
     const manifest = await ManifestEditor.getManifestEditor()
     const { name, vendor, builders } = manifest
     const { dotnet, node, 'service-js': serviceJs } = builders
--- a/src/clients/runtime.ts
+++ b/src/clients/runtime.ts
@@ -55,13 +55,18 @@ export class Runtime {
     const ws = new WebSocket(formattedUrl, clientOptions)
     const wsDuplexStream = (WebSocket as any).createWebSocketStream(ws, { encoding: 'utf8' })
 
+    wsDuplexStream.on('error', () => {
+      console.log('Connection closed')
+      process.exit(0)
+    })
+
+    wsDuplexStream.pipe(process.stdout)
+
     if (interactive) {
       process.stdin.pipe(wsDuplexStream, { end: false })
       process.stdin.on('end', () => {
         wsDuplexStream.end(EOT)
       })
     }
-
-    wsDuplexStream.pipe(process.stdout)
   }
 }
--- a/src/clients/runtime.ts
+++ b/src/clients/runtime.ts
@@ -1,0 +1,67 @@
+import { IOContext } from '@vtex/api'
+import { ManifestEditor } from '../lib/manifest'
+import { cluster } from '../env'
+import * as url from 'url'
+import * as WebSocket from 'ws'
+import { getToken } from '../conf'
+
+const EOT = '\x04'
+
+export class Runtime {
+  private region: string
+  private account: string
+  private workspace: string
+
+  constructor(context: IOContext) {
+    const { region, account, workspace } = context
+    this.region = region
+    this.account = account
+    this.workspace = workspace
+  }
+
+  public executeCommand = async (command: string, interactive: boolean) => {
+    const manifest = await ManifestEditor.getManifestEditor()
+    const { name, vendor, builders } = manifest
+    const { dotnet, node, 'service-js': serviceJs } = builders
+    if (!dotnet && !node && !serviceJs) {
+      return
+    }
+
+    const host = `${name}.${vendor}.${this.region}.vtex.io`
+    const path = `/${this.account}/${this.workspace}/_exec`
+    const clusterHeader = cluster() ? { 'x-vtex-upstream-target': cluster() } : null
+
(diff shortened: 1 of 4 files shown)


src/modules/apps/exec.ts
```diff
@@ -0,0 +1,53 @@
+import * as WebSocket from 'ws'
+import * as url from 'url'
+import { getAccount, getToken, getWorkspace } from '../../conf'
+import { getManifest } from '../../manifest'
+import { region } from '../../env'
+import { toMajorRange } from '../../locator'
+const EOT = '\x04'
+export default async (command: string, options) => {
+  const manifest = await getManifest()
```
tiagonapoli: There's a new `getManifest` function - `await ManifestEditor.getManifestEditor()`

src/modules/apps/exec.ts
```diff
@@ -0,0 +1,53 @@
...
+  const { name, vendor, version, builders } = manifest
+  const { dotnet, node, 'service-js': serviceJs } = builders
+  if (!dotnet && !node && !serviceJs) {
+    return
+  }
+  const majorRange = toMajorRange(version)
```
tiagonapoli: I will add a `majorRange` getter on `ManifestEditor`, then it could be `manifest.majorRange`

tiagonapoli: Added :)

src/modules/apps/exec.ts
```diff
@@ -0,0 +1,53 @@
...
+    hostname: host,
+    pathname: path,
+    query: {
+      __v: majorRange,
+      params: command?.split(' '),
+      interactive: options.i || options.interactive,
+    },
+  }
+  const formattedUrl = url.format(urlObject)
+  const ws = new WebSocket(formattedUrl, clientOptions)
```
tiagonapoli: What do you think on this `WebSocket` being a client on the `clients` folder? I don't know if this would make sense, but would facilitate finding all toolbelt's external communications

src/modules/apps/exec.ts
```diff
@@ -0,0 +1,53 @@
...
+    return
+  }
+  const majorRange = toMajorRange(version)
+  const host = `${name}.${vendor}.${region()}.vtex.io`
+  const path = `/${getAccount()}/${getWorkspace()}/_exec`
+  const clientOptions = {
+    headers: {
+      Authorization: getToken(),
+      Host: host,
+      'X-Vtex-Runtime-Api': 'true',
```
tiagonapoli: I think there's another header that should be added, which specifies the cluster if it's defined:
https://github.com/vtex/toolbelt/blob/3f4941d2edebc2e493367fbea75aae851a2c8c1a/src/clients/index.ts#L30-L37

src/modules/apps/exec.ts
```diff
@@ -0,0 +1,53 @@
...
+    },
+  }
+  var urlObject = {
```
tiagonapoli: You can use `let` or `const` if it's not modified

src/modules/apps/exec.ts
```diff
@@ -0,0 +1,53 @@
...
+      ws.send(EOT)
+    })
+  })
+  ws.on('close', function close() {
+    process.exit(0)
+  })
+  ws.on('message', function incoming(data) {
+    process.stdout.write(data.toString())
+  })
+}
```
tiagonapoli: The handler functions could be arrow function - I think it's a more common pattern

tiagonapoli: How about using the `WebSocket` [Streams API](https://www.npmjs.com/package/ws#use-the-nodejs-streams-api)?
I think this could be:
```ts
const ws = new WebSocket('formattedUrl', clientOptions)
const wsDuplexStream = WebSocket.createWebSocketStream(ws, { encoding: 'utf8' });
process.stdin.pipe(wsDuplexStream)
wsDuplexStream.pipe(process.stdout)
```
I think it would needed to be tested if `EOT` is necessary in this case, otherwise it could be:
```ts
const ws = new WebSocket('formattedUrl', clientOptions)
const wsDuplexStream = WebSocket.createWebSocketStream(ws, { encoding: 'utf8' });
process.stdin.pipe(wsDuplexStream, { end: false })
process.stdin.on('end', () => wsDuplexStream.end(EOT))
wsDuplexStream.pipe(process.stdout)
```

- [pipe docs](https://nodejs.org/api/stream.html#stream_readable_pipe_destination_options)

src/clients/runtime.ts
```diff
@@ -0,0 +1,67 @@
...
+  private workspace: string
+
+  constructor(context: IOContext) {
+    const { region, account, workspace } = context
+    this.region = region
+    this.account = account
+    this.workspace = workspace
+  }
+
+  public executeCommand = async (command: string, interactive: boolean) => {
```
tiagonapoli: In this case you could use `public async executeCommand(...`
When using classes I had doubts on when using named functions `public async executeCommand() {` or arrow functions `public executeCommand = () => {`. The difference in terms of behavior is shown in this example:
```

class A {
  constructor(x) {
    this.x = x
  }

  named() {
    console.log(this.x)
  }

  arrow = () => {
    console.log(this.x)
  }
}

const testArrow = () => {
    const a = new A(300)
 
    // Will print 300
    a.arrow()
    
    const fn1 = a.arrow
    // Will print 300
    fn1()
}

const testNamed = () => {
    const a = new A(400)

    // Will print 400
    a.named()

    const fn1 = a.named
    // Doesn't work - when a.named is decouped from the object 'a' the this variable is decoupled as well, so this.x won't exist 
    fn1() 
}

testArrow()
testNamed()
```

That occurs because when an arrow function is created it binds the content inside it with the current `this` variable, preventing decoupling.

However, creating an arrow function is not so performatic as a named function on a class - so if `runtime.executeCommand` isn't decouple from its object (all calls to it are `runtime.executeCommand(...)`), it's better to use named functions

https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1

Of course in toolbelt we won't have performance issues, but I think maybe it's a good practive to prefer named functions in classes

PR #722, opened 2020-01-27
Fix bugs when cluster is defined

tiagonapoli:
- Fix URL build when cluster is set
- Add cluster upstream header to billing client

```
cd toolbelt && \
git checkout fix/bugs-with-upstream-defined && \
```
Set a cluster to use:
```
vtex config set cluster 'test-3a'
````
Test using `vtex login`.

Check the host of the URL opened to be: `https://$account.$cluster.myvtexdev.com/`

Check `vtex link` if the pod is created on the specified cluster.

#### Types of changes

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -4,22 +4,24 @@ export function colossusEndpoint() {
   return process.env.VTEX_COLOSSUS_ENDPOINT || `https://infra.io.vtex.com/colossus/v0`
 }
 
-export function region(): string {
-  return process.env.VTEX_CLUSTER || Region.Production
-}
-
 export function cluster() {
   return process.env.VTEX_CLUSTER || getCluster() || ''
 }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/env.ts [hunk 1]
```diff
@@ -4,22 +4,24 @@ export function colossusEndpoint() {
...
-export function region(): string {
-  return process.env.VTEX_CLUSTER || Region.Production
-}
-
 export function cluster() {
   return process.env.VTEX_CLUSTER || getCluster() || ''
 }
 
+export function region(): string {
+  return cluster() || Region.Production
```
victorges: Sanity check: Is `''` treated as "falsy"?

tiagonapoli: Yes:
```
node -e "console.log('' == false)"
```

PR #735, opened 2020-02-05
Remove bluebird

tiagonapoli:
Completely remove bluebird.

```
cd toolbelt && \
git checkout chore/partially-remove-bluebird && \
yarn && yarn build && yarn global add file:$PWD
```

#### Types of changes
- [x] Chore

#### Chores checklist
- [x] Update `CHANGELOG.md`


src/modules/rewriter/export.ts
```diff
@@ -61,19 +61,20 @@ const handleExport = async (csvPath: string) => {
...
   })
 
-  await Promise.each(listOfRanges.splice(counter), async ([from, to]) => {
+  for (const [from, to] of listOfRanges.splice(counter)) {
     let result: any
     try {
       result = await rewriter.exportRedirects(from, to)
     } catch (e) {
-      await saveMetainfo(metainfo, EXPORTS, indexHash, counter, listOfRoutes)
+      saveMetainfo(metainfo, EXPORTS, indexHash, counter, listOfRoutes)
```
dhasuda: why you don't need the await here?

src/modules/rewriter/import.ts
```diff
@@ -103,10 +103,15 @@ export default async (csvPath: string, options: any) => {
   if (reset) {
     const indexFiles = await rewriter.routesIndexFiles().then(prop('routeIndexFiles'))
     const indexFileNames = pluck('fileName', indexFiles) || []
-    indexedRoutes = await Promise.mapSeries(indexFileNames, rewriter.routesIndex).then(
-      compose<any, any, any>(pluck('id'), reduce(concat, []))
-    )
+
+    const arr = []
+    for (const el of indexFileNames) {
```
dhasuda: _nit_: change `el` to `name`

src/modules/workspace/promote.ts
```diff
@@ -10,26 +10,25 @@ import useCmd from './use'
...
-        )} workspace\nOnly production workspaces may be promoted\nUse the command ${chalk.blue(
-          'vtex workspace create <workspace> --production'
-        )} to create a production workspace`
-      )
-    }
-  })
+}
 
+const isPromotable = async (workspace: string) => {
+  isMaster(workspace)
```
dhasuda: no need to `await` this?

src/modules/workspace/promote.ts
```diff
@@ -39,11 +38,12 @@ const promptPromoteConfirm = (workspace: string): Promise<any> =>
     }
   )
 
-export default () => {
+export default async () => {
   log.debug('Promoting workspace', currentWorkspace)
-  return isPromotable(currentWorkspace)
+  await isPromotable(currentWorkspace)
     .then(() => promptPromoteConfirm(currentWorkspace))
     .then(() => promote(account, currentWorkspace))
```
dhasuda: Why do you need to await? For me it's odd to use await and also `.then` when you can either:
```
await isPromotable(currentWorkspace)
promptPromoteConfirm(currentWorkspace)
promote(account, currentWorkspace)
```
or
```
return isPromotabel(currentWorkspace).then(...)
```

src/modules/rewriter/export.ts
tiagonapoli: Because `saveMetaInfo` is not an async function nor returns a promise:
```ts
export const saveMetainfo = (metainfo: any, metainfoType: string, fileHash: string, counter: number, data = {}) => {
  if (!metainfo[metainfoType]) {
    metainfo[metainfoType] = {}
  }
  metainfo[metainfoType][fileHash] = { counter, data }
  writeJsonSync(METAINFO_FILE, metainfo, { spaces: 2 })
}
```

src/modules/workspace/promote.ts
tiagonapoli: `isMaster` just throws an error if the workspace is master, it's sync:
```ts
const isMaster = (workspace: string) => {
  if (workspace === 'master') {
    throw new CommandError(`It is not possible to promote workspace ${workspace} to master`)
  }
}
```

PR #751, opened 2020-02-17
Allow testing editions in non-master workspaces

victorges:
This is to allow setting the edition of an account in a separate workspace, something
that was not possible before and thus made testing editions quite cumbersome
(or dangerous).

With this, toolbelt starts calling the new API feature implemented by https://github.com/vtex/tenant-provisioner/pull/14
(this should be the last one to be merged)

Not being able to test an edition before actually setting it on the master workspace of an account.

 - Run `vtex edition set` in a separate workspace
 - Ensure workspace's edition has been set
 - Ensure master's edition is untouched
 - Also test it on `master` just in case

image

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
- - Add "edition app" to vtex init
+ - Add "edition app" to vtex init.
+ - Allow testing editions in non-master workspaces.
 
 ## [2.91.1] - 2020-03-03
 ### Changed
--- a/src/clients/sponsor.ts
+++ b/src/clients/sponsor.ts
@@ -20,11 +20,11 @@ export class Sponsor extends IOClient {
 
   public getEdition = async () => this.http.get(this.routes.getEdition, { metric: 'get-edition' })
 
-  public setEdition = async (account: string, editionApp: string) => {
+  public setEdition = async (account: string, workspace, editionApp: string) => {
     const [edition, version] = editionApp.split('@')
     const [sponsor, editionName] = edition.split('.')
     return this.http.post(
-      this.routes.setEdition(account),
+      this.routes.setEdition(account, workspace),
       { sponsor, edition: editionName, version },
       { metric: 'set-edition' }
     )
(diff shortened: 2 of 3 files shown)


src/modules/sponsor/setEdition.ts
```diff
@@ -22,16 +23,29 @@ const promptSwitchToAccount = async (account: string, initial: boolean) => {
...
   const data = await sponsorClient.getSponsorAccount()
   const sponsorAccount = R.prop('sponsorAccount', data)
+
   if (!sponsorAccount) {
+    if (previousWorkspace !== 'master') {
+      throw new CommandError('Can only set initial edition in master workspace')
+    }
     await promptSwitchToAccount('vtex', true)
-  } else if (previousAccount !== sponsorAccount) {
+  } else {
```
athoscouto: Should we still `promptSwitchToAccount` when `previousAccount == sponsorAccount`?

athoscouto:
LGTM, I'd wait for Napoli or somebody else from dev tools to sign it off

src/clients/sponsor.ts
```diff
@@ -20,11 +20,11 @@ export class Sponsor extends IOClient {
 
   public getEdition = async () => this.http.get(this.routes.getEdition, { metric: 'get-edition' })
 
-  public setEdition = async (account: string, editionApp: string) => {
+  public setEdition = async (account: string, workspace, editionApp: string) => {
```
tiagonapoli: How about:
```suggestion
  public setEdition = async (account: string, workspace: string, editionApp: string) => {
```

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Allow testing editions in non-master workspaces
```
tiagonapoli: _nit_:
```suggestion
- Allow testing editions in non-master workspaces.
```

tiagonapoli:
You can merge without releasing. I will release it in the next batch which will be on 27/2

src/modules/sponsor/setEdition.ts
victorges: This is not a possible scenario, an account cannot be sponsor of itself. The previous check was quite misleading IMO so I decided to remove it.

If you think it's necessary though, I could check if they are the same to explode explicitly here, but I feel this 💥 should be somewhere else, not here (and it's probably already done).

src/modules/sponsor/setEdition.ts
```diff
@@ -45,8 +48,7 @@ export default async (edition: string) => {
     const sponsorClientForSponsorAccount = new Sponsor(getIOContext(), IOClientOptions)
     await sponsorClientForSponsorAccount.setEdition(previousAccount, previousWorkspace, edition)
 
-    const workspaceNotice = previousWorkspace === 'master' ? '' : `in workspace ${chalk.blue(previousWorkspace)} `
-    log.info(`Successfully set edition ${workspaceNotice}of account ${chalk.blue(previousAccount)}.`)
+    log.info(`Successfully set edition${workspaceNotice} of account ${chalk.blue(previousAccount)}.`)
```
tiagonapoli: `_nit_`: maybe should be this? 
```suggestion
    log.info(`Successfully set edition ${workspaceNotice} of account ${chalk.blue(previousAccount)}.`)
```

PR #756, opened 2020-02-22
Add npm publish lock

tiagonapoli:
Add a reminder step when publishing via `npm publish` without the CI.

Screenshot from 2020-02-22 13-58-59

#### Types of changes
- [X] Chore

### DIFF
--- a/scripts/publishLock.sh
+++ b/scripts/publishLock.sh
@@ -1,0 +1,13 @@
+#! /bin/bash
+
+BOLD="\e[1m"
+YELLOW="\e[33m"
+
+if [ ! "$IS_CI" == "true" ]; then
+    echo -e "${BOLD}${YELLOW}"
+    echo -e "============================================================================================="
+    echo -e "The CI is configured to publish when a new tag is created."
+    echo -e "If you still want to publish yourself you have to set the env variable IS_CI to \"true\""
+    echo -e "============================================================================================="
+    exit 1
+fi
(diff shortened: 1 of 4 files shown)


scripts/publishLock.sh
```diff
@@ -0,0 +1,14 @@
+#! /bin/bash
+
+BOLD="\e[1m"
+YELLOW="\e[33m"
+NO_COLOR="\033[0m"
```
dhasuda: Where is it used?

scripts/publishLock.sh
```diff
@@ -0,0 +1,14 @@
...
+
+if [ ! "$IS_CI" == "true" ]; then
+    echo -e "${BOLD}${YELLOW}"
+    echo -e "============================================================================================="
+    echo -e "The CI is configured to publish when a new tag is created."
+    echo -e "If you still want to publish yourself you have to set the env variable IS_CI to \"true\""
```
dhasuda: I don't think it is intuitive the name `IS_CI`. It's perhaps personal, but when I see `IS_CI = true`, the ideia that comes to me is that everything is already integrated and you shouldn't do the deploy yourself

This is _nit_, so don't take it too seriously

scripts/publishLock.sh
tiagonapoli: Oops, it's not used, will remove

scripts/publishLock.sh
tiagonapoli: That's the idea, everything is integrated, so just in case of emergencies one should do manual release

dhasuda: Yeah! But in this emergencies, I think one should set `IS_CI` to false

PR #760, opened 2020-02-28
Fix build event listening

tiagonapoli:
I messed while trying to fix a lint error. 
I think there's no need to update changelog since the bug isn't released.

```
cd toolbelt && \
git checkout fix/event-listening && \
```
Try linking and/or publishing an app

#### Types of changes

### DIFF
--- a/src/modules/build.ts
+++ b/src/modules/build.ts
@@ -60,7 +60,6 @@ const runErrorAction = (code, message, errorActions) => {
 }
 
 const listen = (appOrKey: string, options: ListeningOptions = {}): Promise<Unlisten> => {
-  let unlistenAll
   const listenPromise = new Promise((resolve, reject) => {
     const { waitCompletion, onError = {}, onBuild = false, context = currentContext, senders = null } = options
     const callback = (eventType, eventData) => {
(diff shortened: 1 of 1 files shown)


src/modules/build.ts
```diff
@@ -97,17 +96,13 @@ const listen = (appOrKey: string, options: ListeningOptions = {}): Promise<Unlis
...
     }
+
     const unlisten = onBuildEvent(context, appOrKey, callback, senders)
-    unlistenAll = () => unlisten(...allEvents)
+    const unlistenAll = () => unlisten(...allEvents)
     if (!waitCompletion) {
       resolve(unlistenAll)
     }
-  }).finally(() => {
-    unlistenAll?.()
```
tiagonapoli: No need for this anymore. It was mainly `undefined`

src/modules/build.ts
```diff
@@ -97,17 +96,13 @@ const listen = (appOrKey: string, options: ListeningOptions = {}): Promise<Unlis
...
   }) as Promise<Unlisten>
-
-  listenPromise.catch(() => {}) // eslint-disable-line @typescript-eslint/no-empty-function
```
tiagonapoli: This was added because `finally` was being used, so no need anymore

PR #762, opened 2020-03-02
Remove all templates but `store` for non vtex users

NataliaGodot:
Nowadays `vtex init` is too chaotic. Let's tidy up the house!

For now we will only allow the `store` template for users outside VTEX and also replace the `store-theme` (the current store template) for `store-framework-template`, that is simpler and easier for a quick start.

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/src/modules/init/index.ts
+++ b/src/modules/init/index.ts
@@ -11,17 +11,22 @@ import * as git from './git'
 const VTEXInternalTemplates = [
   // Only show these templates for VTEX e-mail users.
   'graphql-example',
-  'service-example',
-  'react-guide',
-  'masterdata-graphql-guide',
   'payment-provider-example',
+  'admin-example',
+  'delivery-theme',
+  'service-example',
+  'render-guide',
+  'masterdata-graphql-guide',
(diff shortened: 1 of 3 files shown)


tiagonapoli:
The store-framework-template just needs to be migrated to vtex-apps before releasing this PR

src/modules/init/index.ts
```diff
@@ -73,10 +114,10 @@ export default async () => {
   log.debug('Prompting for app info')
   log.info('Hello! I will help you generate basic files and folders for your app.')
   try {
-    const repo = templates[await promptTemplates()]
+    const { repository: repo, organization: org } = templates[await promptTemplates()]
     await promptContinue(repo)
-    log.info(`Cloning https://vtex-apps/${repo}.git`)
-    await git.clone(repo)
+    log.info(`Cloning https://${org}/${repo}.git`)
```
tiagonapoli: Could you remove the `https://` and the `.git`? I think it's non important info for the user, what do you think?

PR #771, opened 2020-03-10
Improvements in toolbelt telemetry

NataliaGodot:
### What is the purpose of this pull request?
- Set an interval to remote flush in telemetry object to `toolbelt-telemetry` app.
- Pass telemetry object to `TelemetryReporter` in a file to avoid having errors since the object can be too large to be a cli argument.
- Send `TelemetryReporter` meta errors to `toolbelt-telemetry`.

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
  - Add "edition app" to vtex init.
  - Allow testing editions in non-master workspaces.
- - Add an interval to remote flush telemetry.
- - Use a file to write telemetry object argument for `ToolbeltReporter`.
+ - Improvements in telemetry.
+  - Add an interval to remote flush telemetry.
+  - Use a file to write telemetry object argument for `TelemetryReporter`.
+  - Send `TelemetryReporter` meta errors to `toolbelt-telemetry`.
 
 ## [2.91.1] - 2020-03-03
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
  - Add "edition app" to vtex init.
  - Allow testing editions in non-master workspaces.
+ - Add an interval to remote flush telemetry.
+ - Use a file to write telemetry object argument for `ToolbeltReporter`.
 
 ## [2.91.1] - 2020-03-03
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Add "edition app" to vtex init.
  - Allow testing editions in non-master workspaces.
  - Improvements in telemetry.
-  - Add an interval to remote flush telemetry.
-  - Use a file to write telemetry object argument for `TelemetryReporter`.
-  - Send `TelemetryReporter` meta errors to `toolbelt-telemetry`.
+    - Add an interval to remote flush telemetry.
+    - Use a file to write telemetry object argument for `TelemetryReporter`.
+    - Send `TelemetryReporter` meta errors to `toolbelt-telemetry`.
 
 ## [2.91.1] - 2020-03-03
 ### Changed
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -179,7 +179,7 @@ const start = async () => {
 
   try {
     await main()
-    TelemetryCollector.getCollector().flush()
+    await TelemetryCollector.getCollector().flush()
   } catch (err) {
     await onError(err)
   }
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -158,7 +158,7 @@ const onError = async (e: any) => {
 
   const errorReport = TelemetryCollector.getCollector().registerError(e)
   log.error(`ErrorID: ${errorReport.errorId}`)
-  TelemetryCollector.getCollector().flush()
+  await TelemetryCollector.getCollector().flush()
   process.exit(1)
 }
 
--- a/src/lib/telemetry/TelemetryCollector.ts
+++ b/src/lib/telemetry/TelemetryCollector.ts
@@ -6,10 +6,11 @@ import { join } from 'path'
 import * as pkgJson from '../../../package.json'
 import { ErrorReport } from '../error/ErrorReport'
 import { ITelemetryLocalStore, TelemetryLocalStore } from './TelemetryStore'
+import { configDir } from '../../conf'
 
 export class TelemetryCollector {
   private static readonly REMOTE_FLUSH_INTERVAL = 1000 * 60 * 10 // Ten minutes
-  private static readonly telemetryObjFilePathPrefix = '~/.config/configstore/vtex/telemetry'
+  private static readonly TELEMETRY_LOCAL_DIR = join(configDir, 'vtex', 'telemetry')
   private static telemetryCollectorSingleton: TelemetryCollector
 
   public static getCollector() {
@@ -66,7 +67,7 @@ export class TelemetryCollector {
       errors: this.errors.map(err => err.toObject()),
       metrics: this.metrics,
     }
-    const objFilePath = `${TelemetryCollector.telemetryObjFilePathPrefix}-${randomBytes(8).toString('hex')}.json`
+    const objFilePath = join(TelemetryCollector.TELEMETRY_LOCAL_DIR, `${randomBytes(8).toString('hex')}.json`)
     try {
       await ensureFile(objFilePath)
       await writeJson(objFilePath, obj) // Telemetry object should be saved in a file since it can be too large to be passed as a cli argument
--- a/src/lib/telemetry/TelemetryCollector.ts
+++ b/src/lib/telemetry/TelemetryCollector.ts
@@ -1,10 +1,15 @@
+import { randomBytes } from 'crypto'
+import { writeJson, ensureFile } from 'fs-extra'
 import { spawn } from 'child_process'
 import { join } from 'path'
+
 import * as pkgJson from '../../../package.json'
 import { ErrorReport } from '../error/ErrorReport'
 import { ITelemetryLocalStore, TelemetryLocalStore } from './TelemetryStore'
 
 export class TelemetryCollector {
+  private static readonly REMOTE_FLUSH_INTERVAL = 1000 * 60 * 10 // Ten minutes
+  private static readonly telemetryObjFilePathPrefix = '~/.config/configstore/vtex/telemetry'
   private static telemetryCollectorSingleton: TelemetryCollector
 
   public static getCollector() {
@@ -43,8 +48,8 @@ export class TelemetryCollector {
 
   public registerMetric() {}
 
-  public flush(forceRemoteFlush = false) {
-    const shouldRemoteFlush = forceRemoteFlush || this.errors.length > 0
+  public async flush(forceRemoteFlush = false) {
+    const shouldRemoteFlush = forceRemoteFlush || this.errors.length > 0 || (Date.now() - this.store.getLastRemoteFlush() >= TelemetryCollector.REMOTE_FLUSH_INTERVAL)
     if (!shouldRemoteFlush) {
       this.store.setErrors(this.errors)
       this.store.setMetrics(this.metrics)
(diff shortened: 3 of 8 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
  - Add "edition app" to vtex init.
  - Allow testing editions in non-master workspaces.
+ - Improvements in telemetry.
+  - Add an interval to remote flush telemetry.
+  - Use a file to write telemetry object argument for `TelemetryReporter`.
+  - Send `TelemetryReporter` meta errors to `toolbelt-telemetry`.
```
tiagonapoli:
```suggestion
 - Improvements in telemetry.
   - Add an interval to remote flush telemetry.
   - Use a file to write telemetry object argument for `TelemetryReporter`.
   - Send `TelemetryReporter` meta errors to `toolbelt-telemetry`.
```
This way the markdown will understand it's a subtopic :)

src/lib/telemetry/TelemetryStore.ts
```diff
@@ -5,6 +5,7 @@ export interface ITelemetryLocalStore {
   storeName: string
   getErrors: () => ErrorReport[]
   getMetrics: () => any
+  getLastRemoteFlush: () => any
```
tiagonapoli: Maybe we should add `setLastRemoteFlush` as well?

src/lib/telemetry/TelemetryCollector.ts
```diff
@@ -1,10 +1,15 @@
...
 import { spawn } from 'child_process'
 import { join } from 'path'
+
 import * as pkgJson from '../../../package.json'
 import { ErrorReport } from '../error/ErrorReport'
 import { ITelemetryLocalStore, TelemetryLocalStore } from './TelemetryStore'
 
 export class TelemetryCollector {
+  private static readonly REMOTE_FLUSH_INTERVAL = 1000 * 60 * 10 // Ten minutes
+  private static readonly telemetryObjFilePathPrefix = '~/.config/configstore/vtex/telemetry'
```
tiagonapoli: Since we are going to use this directory to everything related to telemetry, how about we change the name to something like `LOCAL_TELEMETRY_DIR` or something better?

The `conf.ts` file exports a `configDir`, so this could be:
```ts
private static readonly LOCAL_TELEMETRY_DIR = join(configDir, 'vtex', 'telemetry')
```

tiagonapoli: Just saw that actually we are not using this dir to everything, how about we use? What do you think?
```
vtex/
  +-- telemetry/
    +-- errors/
    +-- hash1.json
    +-- hash2.json
```

src/lib/telemetry/TelemetryCollector.ts
```diff
@@ -58,8 +66,15 @@ export class TelemetryCollector {
       errors: this.errors.map(err => err.toObject()),
       metrics: this.metrics,
     }
+    const objFilePath = `${TelemetryCollector.telemetryObjFilePathPrefix}-${randomBytes(8).toString('hex')}.json`
+    try {
+      await ensureFile(objFilePath)
+      await writeJson(objFilePath, obj) // Telemetry object should be saved in a file since it can be too large to be passed as a cli argument
+    } catch (e) {
+      console.log('Error writing telemetry file. Error: ', e)
```
tiagonapoli: How about using `logger.error`?

src/lib/telemetry/TelemetryCollector.ts
```diff
@@ -58,8 +66,15 @@ export class TelemetryCollector {
...
+    }
 
-    spawn(process.execPath, [join(__dirname, 'TelemetryReporter.js'), this.store.storeName, JSON.stringify(obj)], {
+    spawn(process.execPath, [join(__dirname, 'TelemetryReporter.js'), this.store.storeName, objFilePath], {
```
tiagonapoli: I think this has to be inside the `try` just above, or we could `return` inside the `catch`, because if we didn't succeeded writing the file we shouldn't `spawn` the process

src/lib/telemetry/TelemetryReporter.ts
```diff
@@ -1,13 +1,20 @@
...
 import { createIOContext, createTelemetryClient } from '../clients'
 import { SessionManager } from '../session/SessionManager'
 import { TelemetryLocalStore } from './TelemetryStore'
+import { ErrorReport } from '../error/ErrorReport'
+import { join, resolve } from 'path'
 
 export class TelemetryReporter {
   private static readonly RETRIES = 3
   private static readonly TIMEOUT = 30 * 1000
+  public static readonly ERRORS_DIR = resolve('~/.config/configstore/vtex/errors')
```
tiagonapoli: Hmmmm, I'm wondering about if the errors dir should be an option this script receives, or if it should be:
```ts
public static readonly ERRORS_DIR = join(TelemetryCollector.LOCAL_TELEMETRY_DIR, 'errors')
```
Doing in a way it receives from the arguments and the `TelemetryReporter` class has in its contructor something like:
```
constructor(private errorsDir: string) 
```
Would make testing this class easier, I think, because we could easily mock an `errorDir` that we want for testing.

src/lib/telemetry/TelemetryReporter.ts
```diff
@@ -31,16 +38,58 @@ export class TelemetryReporter {
   }
 }
 
+const lockfilePromisified = (lockName, options) => {
+  return new Promise((resolve, reject) => {
+    lockfile.lock(lockName, options, err => {
+      err ? reject(err) : resolve()
+    })
+  })
+}
```
tiagonapoli: We could create a cool class:
```ts
class FileLock {
  constructor(private lockName: string, private options: any) {}

  public lock() {
      return new Promise((resolve, reject) => {
        lockfile.lock(lockName, options, err => {
        err ? reject(err) : resolve()
      })
  }

  unlock() {
    ...
  }
}
```

src/lib/telemetry/TelemetryReporter.ts
```diff
@@ -31,16 +38,58 @@ export class TelemetryReporter {
...
+        await remove(metaErrorsFile)
+      })
+    )
+    lockfile.unlock(lockfilePath)
+
     store.setLastRemoteFlush(Date.now())
     process.exit()
   } catch (err) {
-    console.log(err)
+    const errorFilePath = `${TelemetryReporter.ERRORS_DIR}/${randomBytes(8).toString('hex')}.json`
```
tiagonapoli: Maybe we should call `unlock` here as well? If we use the `FileLock` class it could store a boolean `isLocked` indicating that the lock was called successfully, then `unlock` would call `lockfile.unlock` only in case `isLocked` is true to avoid errors

tiagonapoli: Actually I think we should call `unlock` at the end of this catch

src/lib/telemetry/TelemetryReporter.ts
tiagonapoli: Actually this wouldn't be necessary, since `TelemetryReporter` is not responsible in the moment for writing errors

tiagonapoli:
I had an idea that could we improve the telemetry reporting workflow and I would like to discuss it:
1. `TelemetryReporter` is called with the format

   ```bash
   node TelemetryReporter.js pathToFileToTryToReport.json pathToPendingDataToSendDirectory 
   ```

2. The `TelemetryReporter` class receives these data
3. It reads the file and removes it
4. Tries to send the data
   1. _Success_: Just goes to step 5
   2. _Failure_: Creates a file inside `pathToPendingDataToSendDirectory` with the error occurred and creates another file with the data it tried to send 
5. Acquires a lock inside `pathToPendingDataToSendDirectory`
6. Tries to send the files located there
7. For each success it removes the file associated with it and for each failure it creates a new file associated with this error
8. Releases lock and terminates

We could improve `TelemetryReporter` to have functions like:

- `private createPendingDataFile`
- `public sendPendingData`
- `public sendTelemetryData`

Then all files inside `pathToPendingDataToSendDirectory` could implement the interface:

```
interface TelemetryData {
  metrics?: any
  error?: object[]
}
```

src/lib/telemetry/TelemetryReporter.ts
```diff
@@ -31,16 +38,58 @@ export class TelemetryReporter {
...
+    let errorReport = err
+    if (!(err instanceof ErrorReport)) {
+      const code = ErrorReport.createGenericCode(err)
+      errorReport = ErrorReport.create({
+        code,
```
tiagonapoli: In this case we could label the code as `TelemtryReporterError` so we could easily find this type of error in splunk :sparkles:

src/lib/telemetry/TelemetryReporter.ts
```diff
@@ -31,16 +38,58 @@ export class TelemetryReporter {
...
+    const metaErrorsFiles = (await readdir(TelemetryReporter.ERRORS_DIR)).filter(fileName => {
+      return fileName !== lockfileName
+    })
+    await Promise.all(
+      metaErrorsFiles.map(async metaErrorsFile => {
+        const metaErrorsObject = await readJson(join(TelemetryReporter.ERRORS_DIR, metaErrorsFile))
+        await reporter.reportErrors(metaErrorsObject)
+        await remove(metaErrorsFile)
+      })
+    )
```
tiagonapoli:
```ts
   let errors: ErrorReport[] = []
    await Promise.all(
      metaErrorsFiles.map(async metaErrorsFile => {
        try {
          const metaErrorsObject = await readJson(join(TelemetryReporter.ERRORS_DIR, metaErrorsFile))
          await reporter.reportErrors(metaErrorsObject)
          await remove(metaErrorsFile)
        } catch(err) {
         .....
         errors.push(newErrorReport)
        }
      })
    )
```

src/cli.ts
```diff
@@ -158,7 +158,7 @@ const onError = async (e: any) => {
 
   const errorReport = TelemetryCollector.getCollector().registerError(e)
   log.error(`ErrorID: ${errorReport.errorId}`)
-  TelemetryCollector.getCollector().flush()
+  await TelemetryCollector.getCollector().flush()
```
tiagonapoli: Maybe it would be ideal to add `await` on the `flush` called down on `start` function?

src/lib/telemetry/TelemetryReporter.ts
```diff
@@ -1,13 +1,47 @@
...
+    return new Promise((resolve, reject) => {
+      lockfile.lock(this.lockPath, this.options, err => {
+        err ? reject(err) : resolve()
+      })
+    })
+  }
+
+  public unlock() {
+    try {
+      lockfile.unlock(this.lockPath)
```
tiagonapoli: I think `lockfile.unlock` works the same way that `lock` works, it receives an callback, so we could wrap it inside a promise:
https://www.npmjs.com/package/lockfile#lockfileunlockpath-cb
```
lockFile.lock(path, [opts], cb)
Acquire a file lock on the specified path

lockFile.lockSync(path, [opts])
Acquire a file lock on the specified path

lockFile.unlock(path, cb)
Close and unlink the lockfile.
```

src/lib/telemetry/TelemetryCollector.ts
```diff
@@ -1,10 +1,17 @@
+import { randomBytes } from 'crypto'
+import { ensureFile, writeJson } from 'fs-extra'
 import { spawn } from 'child_process'
 import { join } from 'path'
+
 import * as pkgJson from '../../../package.json'
 import { ErrorReport } from '../error/ErrorReport'
 import { ITelemetryLocalStore, TelemetryLocalStore } from './TelemetryStore'
+import { configDir } from '../../conf'
+import { logger } from '../../clients'
```
tiagonapoli: Oops, that is not the desired logger, it should be:
```
import logger from '../../logger'
```
That's one of the reasons I don't like `default exports` haha, normally vscode won't show that there's a logger in this file

src/lib/telemetry/TelemetryReporter.ts
```diff
@@ -1,13 +1,47 @@
...
+import { ErrorCodes } from '../error/ErrorCodes'
+
+class FileLock {
+  public readonly lockName: string
+  constructor(private lockPath: string, private options: any) {
+    this.lockName = basename(lockPath)
+  }
+
+  public async lock() {
+    await ensureDir(TelemetryReporter.PENDING_DATA_DIR)
```
tiagonapoli: I think this should be:
```
await ensureDir(dirname(this.lockPath))
```
https://nodejs.org/api/path.html#path_path_dirname_path

src/lib/telemetry/TelemetryReporter.ts
```diff
@@ -21,28 +55,91 @@ export class TelemetryReporter {
...
     )
 
-    return new TelemetryReporter(telemetryClient)
+    return new TelemetryReporter(telemetryClient, TelemetryReporter.PENDING_DATA_DIR)
+  }
+
+  private dataPendingLock: FileLock
+  constructor(private telemetryClient: TelemetryClient, private pendingDataDir: string) {
+    const dataPendingLockName = 'reporter.lock'
+    const dataPendingLockPath = join(TelemetryReporter.PENDING_DATA_DIR, dataPendingLockName)
```
tiagonapoli:
```suggestion
    const dataPendingLockPath = join(this.pendingDataDir, dataPendingLockName)
```

src/lib/telemetry/TelemetryReporter.ts
```diff
@@ -21,28 +55,91 @@ export class TelemetryReporter {
...
+          const pendingDataObject = await readJson(pendingDataFile)
+          await this.reportErrors(pendingDataObject.errors)
+          await remove(pendingDataFile)
+        } catch (err) {
+          errors.push(err)
+        }
+      })
+    )
+
+    errors.length > 0 ?? (await this.createTelemetryReporterMetaError(errors))
```
tiagonapoli: Maybe use `if`?

PR #778, opened 2020-03-16
Cmd logs

daniel-mitre:
Adds a brand new command: `vtex logs`! This allows developers listen the logs from the VTEX IO apps they created (currently only for non-VTEX accounts).

This increases visibility of what happened in their IO apps, helping to visualize the errors this app is throwing in production environments.

1. Switch to an account that uses our logger in their IO apps (`tbb` for example) or use `storecomponents` (you'll probably need to link some application with logs and change the vendor to `storecomponents`).
2. Use `vtex logs --all` and/or `vtex logs --all`
3. Use `vtex logs chaordic-graphql`

If no logs are showing, all logs may already been consumed. Use `vtex logs chaordic-graphql --past` to see logs already seen by someone.

You can also run `vtex logs` in the root path of an IO app and the filter will be read from the manifest.json

#### Types of changes

#### Chores checklist
- [ ] Update `CHANGELOG.md`


daniel-mitre:
I could use some help to fix this data race condition pointed from the linter.

daniel-mitre:
Printing raw objects gives them color highlighting...
image

while JSON.stringify shows intern objects...
image

There's a way to combine them? 😞

tiagonapoli:
You could have asked me directly :(

src/eventsource.ts
```diff
@@ -82,6 +91,9 @@ export default class CustomEventSource {
     this.closeEventSource()
     this.clearTimers()
     this.isClosed = true
+    if (typeof this.esOnClose === 'function') {
+      this.esOnClose()
+    }
```
tiagonapoli: Can this be another thing than `undefined` or `null`? Otherwise this can be:
```suggestion
      this.esOnClose?.()
```

src/modules/tree.ts
```diff
@@ -488,4 +488,23 @@ export default {
       requiredArgs: 'edition',
     },
   },
+  logs: {
+    description: 'Show logs of an app on the current directory or a specified one',
+    handler: './apps/logs',
+    optionalArgs: ['vendor', 'app'],
```
tiagonapoli: Prefer using flags for this - it becomes much clearer for the user, for example:
```
vtex logs --vendor storecomponents
vtex logs --app vtex.builder-hub@0.x
```

src/modules/tree.ts
```diff
@@ -488,4 +488,23 @@ export default {
...
+    options: [
+      {
+        description: 'Show logs of all apps of this vendor',
+        long: 'all',
+        short: 'a',
+        type: 'boolean',
+      },
+      {
+        description: "Show all available logs and don't commit the new consume state",
```
tiagonapoli: Maybe change the description to something simpler to understand?

src/modules/tree.ts
```diff
@@ -488,4 +488,23 @@ export default {
       requiredArgs: 'edition',
     },
   },
+  logs: {
+    description: 'Show logs of an app on the current directory or a specified one',
+    handler: './apps/logs',
+    optionalArgs: ['vendor', 'app'],
+    options: [
+      {
+        description: 'Show logs of all apps of this vendor',
```
tiagonapoli: Seem a bit odd this phrase construction, maybe check with Vladimir a better way to say this?

src/modules/tree.ts
```diff
@@ -488,4 +488,23 @@ export default {
       requiredArgs: 'edition',
     },
   },
+  logs: {
+    description: 'Show logs of an app on the current directory or a specified one',
```
tiagonapoli: `Show logs from an app` or `Show logs of an app`? :thinking:

src/modules/apps/logs.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export default async (vendor: string, app: string, options) => {
+  const account = getAccount()
+  const workspace = 'master'
+  const skidderMajor = 1
+
+  const conf = {
+    headers: {
+      Authorization: `${AuthType.bearer} ${getToken()}`,
+      'user-agent': `${userAgent}${options.ghost ? `#${Math.random()}` : ''}`,
```
tiagonapoli: Use `randomBytes(8).toString('hex')` instead of `Math.random()`

src/modules/apps/logs.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    es.addEventListener('message', msg => {
+      console.log(JSON.stringify(JSON.parse(msg.data).data, null, 2))
+    })
+  }
+
+  createEventSource()
+
+  console.log('Press CTRL+C to abort')
```
tiagonapoli: Use `log.debug`?

src/modules/apps/logs.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    es.onclose = () => {
+      log.warn('SSE connection closed. Reconnecting.')
+      createEventSource()
+    }
+
+    es.addEventListener('message', msg => {
+      console.log(JSON.stringify(JSON.parse(msg.data).data, null, 2))
+    })
+  }
```
tiagonapoli: I think it would be clearer if this function stayed out of this and received the necessary arguments

src/modules/apps/logs.ts
```diff
@@ -0,0 +1,73 @@
...
+  if (options.all) {
+    app = ''
+  }
+
+  let uri = `http://infra.io.vtex.com/skidder/v${skidderMajor}/${vendor}/${workspace}/logs/stream`
+  if (app) {
+    uri += `/${app}`
+  }
+
+  function createEventSource() {
```
tiagonapoli: Maybe another name? `createLogEventSource`?

src/modules/apps/logs.ts
```diff
@@ -0,0 +1,73 @@
...
+    const es = new CustomEventSource(uri, conf)
+    console.info(`Listening ${vendor}${app ? `.${app}` : ''} logs`)
```
tiagonapoli: `logger.info`

src/modules/apps/logs.ts
```diff
@@ -0,0 +1,73 @@
...
+      log.error(`Error reading logs: ${err.message}`)
+    }
+
+    es.onclose = () => {
+      log.warn('SSE connection closed. Reconnecting.')
+      createEventSource()
+    }
+
+    es.addEventListener('message', msg => {
+      console.log(JSON.stringify(JSON.parse(msg.data).data, null, 2))
```
tiagonapoli: Use `util.inspect` for showing pretty logs - you can specify the depth desired: https://nodejs.org/api/util.html#util_util_inspect_object_options

src/modules/apps/logs.ts
```diff
@@ -0,0 +1,73 @@
...
+import { getManifest } from '../../manifest'
+import CustomEventSource from '../../eventsource'
+import userAgent from '../../user-agent'
+import log from '../../logger'
+import { AuthType } from '@vtex/api'
+
+export default async (vendor: string, app: string, options) => {
+  const account = getAccount()
+  const workspace = 'master'
+  const skidderMajor = 1
```
tiagonapoli: Maybe a constant in the beggining of the file?
```ts
const SKIDDER_MAJOR = 1
```

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- [vtex logs] Add logs command
```
tiagonapoli, line 9: _nit_:
```suggestion
- [vtex logs] Add logs command.
```

src/lib/sse/CustomEventSource.ts
```diff
@@ -89,6 +90,14 @@ export class CustomEventSource {
     }
   }
 
+  set onclose(newOnClose) {
+    this.esOnClose = newOnClose
+    this.esOnClose = this.esOnClose.bind(this)
+    if (this.eventSource) {
+      (this.eventSource as any).onclose = this.esOnClose
+    }
```
tiagonapoli, line 98: O think the `eventSource` doesn't have a `onclose` method
https://github.com/EventSource/eventsource/blob/61e1b19c8616aa151835a4ae599b299afb574ebf/lib/eventsource.js#L317

src/modules/apps/logs.ts
```diff
@@ -0,0 +1,79 @@
...
+    const streamLocator = `${account}${app ? `.${app}` : ''}`
+    log.info(`Connecting to logs stream for ${streamLocator}`)
+    log.debug(`Stream URI ${uri}`)
+
+    es.onopen = () => {
+      log.info(`Listening to ${streamLocator}'s logs`)
+    }
+
+    es.onerror = err => {
+      log.error(`Error reading logs: ${err.message}`)
```
tiagonapoli, line 59: Maybe it's in our interest to add this to telemetry?

You could add to `ErrorKinds` something like `APP_LOGS_SSE_ERROR:  'SSEError:AppLogs'`, or just use the `SSEError` kind and here:
```ts
ErrorReport.createAndRegisterOnTelemetry({
  kind: ErrorKinds.SSE_ERROR,
  originalError: err,
  details: {
    sseType: 'logs'
  }
})
```
or
```ts
ErrorReport.createAndRegisterOnTelemetry({
  kind: ErrorKinds.APP_LOGS_SSE_ERROR,
  originalError: err,
})
```

I don't know which would be better though

src/modules/apps/logs.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    es.onclose = () => {
+      log.warn('SSE connection closed. Reconnecting.')
+      createLogEventSource()
+    }
```
tiagonapoli, line 65: I think this is not necessary - the underlying eventSource implementation already deals with reconnection, so the `onclose` addition on `CustomEventSource` can be removed as well

src/lib/sse/CustomEventSource.ts
Jeymisson: sure!

src/modules/apps/logs.ts
Jeymisson: This would be nice! I've added for parsing errors as well

PR #781, opened 2020-03-17
Support debugging .Net apps 

rafaelbac:
To give developers a better experience, we are adding support for debugging .Net applications.

Develop applications without debugging is hard and time-consuming.  

1. Link a dotnet 2.x app (.Net Core 3.0)
2. Add a breakpoint to your code and call the corresponding endpoint
3. Check if the execution stopped in the breakpoint as expected

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Support for debugging .Net Core apps
```
tiagonapoli:
```suggestion
- Support for debugging .Net Core apps.
```

src/modules/debug/dotnet.ts
```diff
@@ -0,0 +1,7 @@
+import { Runtime } from '../../clients/runtime'
+import { getIOContext } from '../utils'
+
+export default async (debugInst: string) => {
+  const runtimeClient = new Runtime(getIOContext())
+  await runtimeClient.debugDotnetApp(debugInst)
```
tiagonapoli: How about this instance function receives a manifest as argument, then the manifest is read before this line and we can warn the user that this command can't be used if dotnet builder is not used:
```ts
  const manifest = await ManifestEditor.getManifestEditor()
  const { builders } = manifest
  if(!builders?.dotnet) {
    logger.error('This command can only be used for dotnet apps')
    process.exit()
  }

  await runtimeClient.debugDotnetApp(debugInst, manifest)
```

src/clients/runtime.ts
```diff
@@ -0,0 +1,69 @@
...
+      },
+    }
+
+    const urlObject = {
+      protocol: 'ws',
+      hostname: host,
+      pathname: path,
+      query: {
+        __v: manifest.majorRange,
+        inst: debugInst?.split(' '),
```
tiagonapoli: Can `debugInst` be null?

src/clients/runtime.ts
```diff
@@ -0,0 +1,69 @@
...
+      },
+    }
+    const formattedUrl = url.format(urlObject)
+
+    const ws = new WebSocket(formattedUrl, clientOptions)
+    const wsDuplexStream = (WebSocket as any).createWebSocketStream(ws, { encoding: 'utf8' })
+
+    wsDuplexStream.on('error', () => {
+      console.log('Connection closed')
```
tiagonapoli: Maybe use `logger.debug`, from `logger.ts` file

src/modules/debug/dotnet.ts
rafaelbac: I already do this check at `runtimeClient.debugDotnetApp()`, but without printing any kind of error message.  Do you want me to move it to the module instead?

src/clients/runtime.ts
rafaelbac: No, it doesn't. Now the parameter is mandatory. I forgot to remove it. Nice catch!

src/modules/debug/dotnet.ts
tiagonapoli: I think it would be better - the client seems a little odd to me to log errors to the stdout - the suggestion to move the manifest instantiation would prevent doing the same work two times

src/modules/debug/dotnet.ts
```diff
@@ -0,0 +1,17 @@
+import { Runtime } from '../../clients/runtime'
+import { getIOContext } from '../utils'
+import { ManifestEditor } from '../../lib/manifest'
+import logger from '../../logger'
+
+export default async (debugInst: string) => {
+  const manifest = await ManifestEditor.getManifestEditor()
+  const { name, vendor, builders } = manifest
+  const { dotnet } = builders
+  if (!dotnet) {
```
tiagonapoli: _nit_: This would be super rare, but in case `builders` is not defined in the manifest:
```suggestion
  if (!builders?.dotnet) {
```

PR #785, opened 2020-03-23
Fix payload too large

NataliaGodot:
Fix PayloadTooLarge error by:
- Truncate strings when necessary
- Set encoding-type in request
- Use a buffer for request body

PayloadTooLarge error.

Go to account `storecomponents`, workspace `toolbelttest` where the [necessary version](https://github.com/vtex/toolbelt-telemetry/pull/6) of `toolbelt-telemetry` is linked. 

Then clone `toolbelt` and run `yarn watch` in this branch with the commands bellow:
```
cd toolbelt && \
git checkout fix/deprecation-check && \
yarn && yarn watch
```

Throw an error with a large message in a command file. For example, in `src/modules/apps/list.ts` in beginning of the default export put the line:
```
throw new Error('a'.repeat(2 ** 20))
```

The use:
```
vtex-test ls
```
To run the command that will thrown the error. After some time, your error should appear in splunk. (Try the query: `index=io_vtex_logs app="vtex.toolbelt-telemetry@*" level=error workspace=toolbelttest`)

#### Types of changes

#### Chores checklist
- [ ] Update `CHANGELOG.md`


src/lib/error/ErrorReport.ts
```diff
@@ -131,16 +135,25 @@ export class ErrorReport extends Error {
...
       kind: this.kind,
-      message: this.message,
+      message: this.maybeTruncateField(this.message),
       errorDetails: this.errorDetails,
-      stack: this.stack,
+      stack: this.maybeTruncateField(this.stack),
       env: this.env,
       ...(this.originalError.code ? { code: this.originalError.code } : null),
     }
+    console.log(r)
```
tiagonapoli: This can be set to production mode again (removing the `r` variable and the `console.log`)

src/lib/error/ErrorReport.ts
```diff
@@ -140,7 +144,29 @@ export class ErrorReport extends Error {
...
+        element[key] = this.truncateStringsFromObject(element[key], maxStrSize)
+      })
+      return element
+    }
+    if(Array.isArray(element)) {
+      element.forEach((elementFromArray, index) => {
+        element[index] = this.truncateStringsFromObject(elementFromArray, maxStrSize)
+      })
+      return element
+    }
```
tiagonapoli: This has to be before `typeof element === 'object'`, because array is considered to be an object

src/lib/error/ErrorReport.ts
```diff
@@ -140,7 +144,29 @@ export class ErrorReport extends Error {
       stack: this.stack,
       env: this.env,
       ...(this.originalError.code ? { code: this.originalError.code } : null),
+    })
+  }
+
+  private truncateStringsFromObject(element: any, maxStrSize: number = ErrorReport.MAX_ERROR_STRING_LENGTH) {
```
tiagonapoli: This can be a `public static` so we can create tests for it in the future - or even a new file exporting this function, what do you think?

src/lib/error/ErrorReport.ts
```diff
@@ -140,7 +144,29 @@ export class ErrorReport extends Error {
...
+    if(element === null || element === undefined) {
```
tiagonapoli: This can be `element == null` - it checks both when `==` is used instead of `===`

tiagonapoli:
Oh, just a detail - ci/lint is failing

src/lib/error/utils.test.ts
```diff
@@ -0,0 +1,17 @@
...
+  [undefined, 10],
+  [null, 10],
+  [{ a: undefined, b: null }, 10],
+  [
+    {
+      a: [{ aa: '123456' }, { ab: '1234' }, { ac: '12345' }],
+      b: { ba: undefined, bb: null, bc: '123456', bd: { bca: '123456', bcb: '1234567' } },
+    },
+  ],
+])('Works on testcase %# - ', (obj: any, limit: number) => {
```
tiagonapoli: If we change the order on `obj: any, limit: number`, we can change the text to:
```
'Works on testcase %# - String limit %d'
```
So we can check the snapshots easier

PR #796, opened 2020-03-28
Add init time metric

NataliaGodot:
Add time to init as a metric for toolbelt telemetry.

Lack of visibility in latency to initialize.

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -41,7 +41,7 @@ const checkLogin = args => {
   }
 }
 
-const main = async () => {
+const main = async (initTimeStartTime?: [number, number]) => {
   const args = process.argv.slice(2)
   conf.saveEnvironment(conf.Environment.Production) // Just to be backwards compatible with who used staging previously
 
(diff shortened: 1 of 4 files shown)


src/cli.ts
```diff
@@ -180,13 +189,14 @@ axios.interceptors.request.use(config => {
 process.on('unhandledRejection', onError)
 
 const start = async () => {
+  const initTimeStartTime = process.hrtime()
```
tiagonapoli: This should be the first thing in this file, so we include the time to import all libs.

This way it wouldn't be necessary to modify `main`'s arguments

tiagonapoli: Just after: `#!/usr/bin/env node`

src/cli.ts
```diff
@@ -62,14 +62,23 @@ const main = async () => {
 
   await checkAndOpenNPSLink()
 
+  if (initTimeStartTime) {
+    const initTime = process.hrtime(initTimeStartTime)
+    const initTimeMetric: Metric = {
+      command: command.command.alias || command.command.path,
+      initTime: 1e3 * initTime[0] + initTime[1] / 1e6,
```
tiagonapoli: You could create on `utils` a function:
```ts
export function hrTimeToMs(hrtime: [number, number]) {
  return 1e3 * hrtime[0] + hrtime[1] / 1e6
}
```

PR #812, opened 2020-04-06
Convert remaining HTTP calls to HTTPS

igorframos:
We must not use HTTP from toolbelt. This PR converts the last of those into HTTPS by calling the new host. HTTP calls to `*.{{region}}.vtex.io` are deprecated and are soon going to stop working, which is also another reason to make this change now.

Another issue that is being fixed by this PR is the fact that we were not encrypting our websockets. Instead of `ws`, we should use `wss`. While this was a trickier change, as Kube Router had a bug that would not allow us to use the new URLs, this bug is fixed in `v6.2.3` and this can be updated here now.

There should be no other instances of `*.{{region}}.vtex.io` in this repo. I have looked for them and I believe I have updated the host everywhere. If any remains, please let me know and/or fix it.

### DIFF
--- a/src/modules/apps/debugger.ts
+++ b/src/modules/apps/debugger.ts
@@ -2,9 +2,9 @@ import streamToString from 'get-stream'
 import net from 'net'
 import WebSocket from 'ws'
 import { getAccount, getToken, getWorkspace } from '../../conf'
-import { region } from '../../env'
 import { ManifestEditor } from '../../lib/manifest'
-import { toMajorRange } from '../../locator'
+import { versionMajor } from '../../locator'
+import { cluster } from '../../env'
 import log from '../../logger'
 
 const keepAliveDelayMs = 3 * 60 * 1000
@@ -28,8 +28,9 @@ function webSocketTunnelHandler(host, path: string): (socket: net.Socket) => voi
   const options = {
     headers: {
       Authorization: getToken(),
-      Host: host,
+      Host: `app.io.vtex.com`,
       'X-Vtex-Runtime-Api': 'true',
+	  'x-vtex-upstream-target': cluster(),
     },
   }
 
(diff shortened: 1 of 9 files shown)


src/modules/utils.ts
```diff
@@ -35,6 +35,9 @@ const DEFAULT_TIMEOUT = 10000
 export const IOClientOptions = {
   timeout: (envTimeout || DEFAULT_TIMEOUT) as number,
   retries: 3,
+  headers: {
+    'x-vtex-upstream-target': env.cluster(),
```
tiagonapoli, line 39: I think this header should be added only if `env.cluster()` is defined
```suggestion
    ...env.cluster() ? { 'x-vtex-upstream-target': env.cluster() } : null
```

src/modules/support/login.ts
```diff
@@ -7,13 +7,14 @@ import { getAccount, getToken, getWorkspace, saveAccount, saveToken, saveWorkspa
...
-const getAvailableRoles = async (region: string, token: string, supportedAccount: string): Promise<string[]> => {
+const getAvailableRoles = async (token: string, supportedAccount: string): Promise<string[]> => {
   const response = await axios.get(
-    `http://support-authority.vtex.${region}.vtex.io/${getAccount()}/${getWorkspace()}/${supportedAccount}/roles`,
+    `https://app.io.vtex.com/vtex.support-authority/v0/${getAccount()}/${getWorkspace()}/${supportedAccount}/roles`,
     {
       headers: {
         Authorization: token,
         'X-Vtex-Original-Credential': token,
+        'x-vtex-upstream-target': env.cluster(),
```
tiagonapoli, line 17:
```suggestion
    ...env.cluster() ? { 'x-vtex-upstream-target': env.cluster() } : null,
```

src/modules/apps/debugger.ts
```diff
@@ -28,14 +28,15 @@ function webSocketTunnelHandler(host, path: string): (socket: net.Socket) => voi
   const options = {
     headers: {
       Authorization: getToken(),
-      Host: host,
+      Host: `app.io.vtex.com`,
       'X-Vtex-Runtime-Api': 'true',
+      'x-vtex-upstream-target': cluster(),
```
tiagonapoli, line 33:
```suggestion
    ...env.cluster() ? { 'x-vtex-upstream-target': env.cluster() } : null,
```

PR #814, opened 2020-04-07
Add support for generating GraphQL types

lucasecdb:

Adds support for generating TypeScript types out of GraphQL operations (queries and mutations) with a new flag in the setup process: `vtex setup --graphql`.

The way I solved this was to reutilize the `apollo` CLI package (but the solution isn't tied to Apollo itself, only the CLI), as it already had a function to generate a TypeScript file out of a GraphQL document (which you can think of as the AST of the GraphQL _file_) and with the server schema.

First, I needed to figure out which apps in the current app `manifest.json` are indeed GraphQL servers, so I iterated each one of them and queried the registry to see if they have the `graphql` builder in their manifests. I included the own app if it has a `graphql` builder because they could query itself.

Then, I queried all GraphQL servers with the [introspection query](https://github.com/graphql/graphql-js/blob/master/src/utilities/getIntrospectionQuery.js#L19-L130) to retrieve all apps schemas, and merged them together. The only thing left to do is get all GraphQL files, parse them, and pass the result to the generate function of `apollo`.

I originally created a thread in Slack asking for this feature:

> I wanted to have strict type definitions for my queries and mutations. Currently, I'm reusing the same type exported by the GraphQL app, but the query or mutation doesn't map one-to-one with the type (you can specify a subset for example), and having generated TypeScript types for them automatically would be awesome.

> The Apollo CLI already does this, but you need to specify the endpoint for your GraphQL server, and it stores the Schema Instrospection result inside the repository in the form of a JSON.

This pull request aims to solve the problem above.

You can access any React app that uses some GraphQL query (the only builder supported for now) and run `vtex-test setup --graphql`. Toolbelt should be able to correctly generate the types for the queries and mutations based on the GraphQL servers specified in the dependencies of `manifest.json`.

Apps you can use to test:

- [test-app](https://github.com/lucasecdb/test-app)
  Simple test app, has one query and one mutation.
- [checkout-resources](https://github.com/vtex-apps/checkout-resources)
  A bit more complicated app, which uses fragments imports and consumes from multiple GraphQL servers for different queries.
- [admin-search](https://github.com/vtex/admin-search)
  App which is self-consuming, i.e. the GraphQL server and consumer exist in the same app.

Result of running `vtex setup --graphql` in `checkout-resources`:

image

#### Types of changes

#### Chores checklist
- [ ] Update `CHANGELOG.md`


lucasecdb:
We decided to not follow with this approach because we don't want toolbelt to worry about compiling GraphQL files to TypeScript (should probably move to another CLI or to `builder-hub`).

PR #834, opened 2020-04-23
Auto-switch to edition-store when creating new workspace

victorges:
This is to reduce the noise about our current editions, by creating some
logic in toolbelt to automatically switch to the `edition-store` when the
current account is in the default edition `edition-business`.

The whole context is: 
 - Stores have to be created with the initial edition `edition-business`, as we don't have any
information on whether it is going to be implemented in IO or Portal, and we don't want to
affect the provision experience of accounts running in the legacy/current platform. 
 - If it is going to be developed in IO though, some user in the account will necessarily need to
create a workspace on it for development/test purposes.
 - With that, we got to the realisation that the moment the user creates a workspace is the
ideal time to prompt them to switch their account's edition.

This basically implements that logic, by prompting the user to switch the edition when
they either create or reset a workspacce.

I had implemented it initially on the `vtex use` command, and only later realised that it
should be in the create/reset commands instead. I still kept the whole commit history as
I think the iteration still makes sense for the reviewer, so I think it whole change is still
good for being reviewed commit-by-commit (if the reviewer likes that 😛). 

Trying to amend the confusion about our editions and when to switch from the default
edition (business) to the one for stores.

 - `vtex workspace create/reset` should now prompt for switching the account edition if it is in the `edition-business` (e.g. `basedevmkp`)
 - `vtex use` should give that prompt only when the workspace is being created or reset
 - Accounts not in `edition-business` (or without any edition) should not be affected by this
 - Ensure user always end up in a valid context (e.g. should not get stuck in the sponsor account after the update) 
   * For that, play around with the prompts by testing answering `no` to some of them.

Screen Shot 2020-04-23 at 20 11 02

#### Types of change

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Update release notes message.
 - Improve SSE errors logs.
 - [vtex setup] Add GraphQL files to `lint-staged` config.
+- [vtex workspace] Auto-switch to edition-store when creating new workspace
 
 ## [2.98.0] - 2020-04-22
 ### Added
@@ -47,7 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [clients:sponsor, vtex support] Use HTTPS and new URL `app.io.vtex.com` format.
 - [telemetry:collector] Flush just before exiting process.
 - [sse] Check if token is valid before creating connection.
-- [sse] Abort process on 401 or 403 server errors. 
+- [sse] Abort process on 401 or 403 server errors.
 - [sse:telemetry] Register sse errors with errorKind `SSEError`.
 - Update release notes message.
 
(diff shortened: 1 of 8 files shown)


victorges:
Reminder for mysefl: Still need to find an actual documentation about the store edition to link in the prompt message for switching.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- [vtex workspace] Auto-switch to edition-store when creating new workspace
```
tiagonapoli, line 8: _nit_:
```suggestion
- [vtex workspace] Auto-switch to edition-store when creating new workspace.
```

src/modules/workspace/common/edition.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+const promptSwitchEdition = (currEditionId: string) => {
+  log.warn(`This account is using the edition ${chalk.blue(currEditionId)}.`)
+  log.warn(
+    `If you are developing your store in IO, it is strongly recommended that you switch to the ${chalk.blue(
+      recommendedEdition
+    )}.`
+  )
+  log.warn(`For more information, visit ${chalk.blue('https://some.doc/about/edition/store')}`)
```
tiagonapoli, line 26: Just a reminder of your reminder to self

PR #835, opened 2020-04-24
Improve SSE errors

tiagonapoli:
This PR adds more info on `SSEErrors` so we can investigate further.
Also I did some minor refactorings.

```
cd toolbelt && \
git checkout fix/sse-errors && \
yarn && yarn watch
```
To test if the SSE is still working properly:
- Try logging in.
- Try linking an app and check if logs are still reaching toolbelt.

SSEError example: https://splunk72.vtex.com/en-US/app/vtex_io_apps/search?earliest=-4h%40m&latest=now&q=search%20index%3Dio_vtex_logs%20app%3Dvtex.toolbelt-telemetry%40*%200542e715c4b40d48&display.page.search.mode=fast&dispatch.sample_ratio=1&display.general.type=events&sid=1587746004.644090_3DDE77BC-1901-4189-B912-ADE751E100F6&workload_pool=

#### Types of changes
- [x] Refactor (non-breaking change that only makes the code better)

#### Chores checklist
- [x] Update `CHANGELOG.md`


src/lib/error/ErrorReport.ts
```diff
@@ -114,11 +115,7 @@ export class ErrorReport extends Error {
...
   }
 
-  private static getRequestErrorMetadata(err: AxiosError): RequestErrorDetails | null {
-    if (!err.config) {
-      return null
-    }
-
+  private static parseAxiosError(err: AxiosError): RequestErrorDetails {
     const { url, method, headers: requestHeaders, params, data: requestData, timeout: requestTimeout } = err.config
     const { status, statusText, headers: responseHeaders, data: responseData } = err.response || {}
```
NataliaGodot, line 120: Couldn't it generate an error if `err.response` is undefined? Since the check to call this function will be true if only `err.config` is defined.

src/lib/sse/CustomEventSource.ts
```diff
@@ -11,13 +14,40 @@ const CONNECTION_CLOSED = 2
...
+    } else {
+      token = SessionManager.getSessionManager().token
+    }
+
+    return new CustomEventSource(source, {
+      headers: {
+        authorization: `bearer ${token}`,
+        'user-agent': userAgent,
+        ...(envCookies() ? { cookie: envCookies() } : null),
+        ...(cluster() ? { 'x-vtex-upstream-target': cluster() } : null),
```
NataliaGodot, line 40: Just an idea as an improvement (not necessary for this PR): we could make a const object `VTEX_HEADERS` and save the headers as constants there. Like this:
```
const VTEX_HEADERS = {
   UPSTREAM_TARGET = 'x-vtex-upstream-target',
   ...
}
```

src/lib/sse/CustomEventSource.ts
```diff
@@ -43,26 +77,27 @@ export default class CustomEventSource {
 
     this.connectEventSource()
     this.addColossusPing()
-    this.reconnectInterval = (this.eventSource && this.eventSource.reconnectInterval) || DEFAULT_RECONNECT_INTERVAL
+    this.reconnectInterval =
+      (this.eventSource && (this.eventSource as any).reconnectInterval) || DEFAULT_RECONNECT_INTERVAL
```
NataliaGodot, line 81:
```suggestion
      (this.eventSource?.reconnectInterval) || DEFAULT_RECONNECT_INTERVAL
```

src/lib/sse/EventSourceError.ts
```diff
@@ -0,0 +1,26 @@
...
+  config: any
+}
+
+export class EventSourceError extends Error {
+  public event: any
+  public eventSourceInfo: EventSourceInfo
+  constructor(event: any, eventSourceInfo: EventSourceInfo) {
+    super(`SSE error on endpoint ${eventSourceInfo.url}`)
+    this.eventSourceInfo = eventSourceInfo
+    this.event = { ...event }
```
NataliaGodot, line 17: Why copying the event?

src/lib/error/ErrorReport.ts
tiagonapoli: Axios requests leave `err.response` undefined when there was a client error on the request (ECONNABORTED, ETIMEOUT, etc), so `err.response == undefined` is a valid request error scenario (guaranteed, I think, to be a client error)

src/lib/sse/CustomEventSource.ts
tiagonapoli: Really nice idea!!!! In the future we definitely need this

src/lib/sse/EventSourceError.ts
tiagonapoli: This was because of an error I got - the `event` object created by the `eventsource` lib has some write protections. This created an error when calling `ErrorReport.toObject` because when truncating the strings from the object at some point we do `object[key] = ...` (this is forbidden because of the write protection)

One way to do this write protection for example is using: https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze

NataliaGodot: Nice!!

PR #836, opened 2020-04-24
Add DEBUG_CP env var for debugging child_processes

tiagonapoli:
Just as the title says

`yarn watch` this version and run commands like `DEBUG_CP=* vtex-test workspace reset`
(I added `export DEBUG_CP=*` on my `.bashrc` to always see child processes output)

#### Types of changes

#### Chores checklist
- [X] Update `CHANGELOG.md`


src/lib/constants/EnvVariablesConstants.ts
```diff
@@ -0,0 +1,3 @@
+export enum EnvVariablesConstants {
```
NataliaGodot, line 1: NIT:
What do you think about `src/lib/constants/EnvVariables.ts` instead? That is, changing the file name from `EnvVariableContants.ts` to `EnvVariables.ts`. (And the same for `PathConstants.ts` in te same folder)

PR #838, opened 2020-04-24
Customize help format

NataliaGodot:
Allow to customize help format. Add a blank line between command and description in help.

Improving help format.

With this branch, just use `--help` with any command and see if there is a blank line between the command and its description.

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/src/commands/switch.ts
+++ b/src/commands/switch.ts
@@ -6,7 +6,7 @@ import authSwitch from '../modules/auth/switch'
 export default class Switch extends CustomCommand {
   static description = 'Switch to another VTEX account'
 
-  static examples = ['vtex switch storecomponents', 'vtex switch storecomponents/myworkspace']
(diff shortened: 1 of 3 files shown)


src/commands/switch.ts
```diff
@@ -6,7 +6,7 @@ import authSwitch from '../modules/auth/switch'
 export default class Switch extends CustomCommand {
   static description = 'Switch to another VTEX account'
 
-  static examples = ['vtex switch storecomponents', 'vtex switch storecomponents myworkspace']
+  static examples = ['vtex switch storecomponents', 'vtex switch storecomponents/myworkspace']
```
tiagonapoli, line 9: I think this syntax of using `account/workspace` is kinda shadowy - an argument is becoming two 
I think we should document on examples the way using flags: `vtex switch storecomponents -w myWorkspace`

PR #839, opened 2020-04-24
Fix install when user cancel command

NataliaGodot:
When we killed `UserCancelledError`, we accidentally inserted a bug in the install command. When the user answered `no` to a prompt in the install command, the command continued.

image

This PR fix this bug.

Try to install an app in the master workspace of an account. A prompt will be showed asking if you are sure. Answer no to the prompt and check if the app was indeed **not installed**.

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Fixed
 - [redirects] Encode delimiter before exporing `.csv` file to avoid conflicts.
+- [vtex install] Behaviour when action is cancelled by the user.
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Fixed
 - [redirects] Encode delimiter before exporing `.csv` file to avoid conflicts.
+- [vtex install] Behaviour when action is cancelled by the user.
```
tiagonapoli, line 10: Maybe [vtex install, vtex uninstall]?

PR #840, opened 2020-04-28
Save workspace promotion info

NataliaGodot:
Now evolution manager have an api to save info about workspace promotions (to allow some audit). This PR makes toolbelt call this api when a `promote` command is used successfully.

Lack of audibility in workspace promotions.

This is a bit complicated since you will need to promote a workspace. My suggestion is to create a production workspace in a test account like `basedevmkp` and immediately promote it. Wait some seconds, then go to GraphiQL in that account admin, select `evolution-manager-graphql` as your app and try the query:
```
query getPromotions {
  promotions { account, agent, workspace, createdIn }
}
``` 
The result should show your workspace promotion.

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/src/clients/evolutionManager.ts
+++ b/src/clients/evolutionManager.ts
@@ -6,8 +6,8 @@ export class EvolutionManager extends AppGraphQLClient {
     super('vtex.evolution-manager-graphql@0.x', context, options)
   }
 
-  public saveWorkspacePromotion = (user: string, workspace: string): Promise<boolean> =>
-    this.graphql
+  public saveWorkspacePromotion(user: string, workspace: string): Promise<boolean> {
+    return this.graphql
       .mutate<{ saveWorkspacePromotion: boolean }, { user: string; workspace: string }>(
         {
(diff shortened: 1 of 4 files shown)


src/clients/evolutionManager.ts
```diff
@@ -0,0 +1,35 @@
+import { InstanceOptions, IOContext, AppGraphQLClient } from '@vtex/api'
+import { GraphQlError } from '../errors'
+
+export class EvolutionManager extends AppGraphQLClient {
+  constructor(context: IOContext, options: InstanceOptions) {
+    super('vtex.evolution-manager-graphql@0.x', context, options)
+  }
+
+  public saveWorkspacePromotion = (user: string, workspace: string): Promise<boolean> =>
```
tiagonapoli, line 9: _nit_: Using arrow function in this case is unnecessary
```suggestion
  public saveWorkspacePromotion(user: string, workspace: string): Promise<boolean>
```

PR #846, opened 2020-04-30
Fix SessionManager and account switches

tiagonapoli:
- Use only SessionManager for managing session.
- Remove session related function on conf.
- Create unified function for switching back to previous account
- Remove unnused code
- Refactor `vtex login`
- Do not report `CommandError`.

```
git checkout fix/session-management

yarn watch
```
- Check if `vtex-test login` is working properly - try logging in to an account and check `~/.vtex/session/session.json` and `~/.vtex/session/workspace.json`.
- Check if `vtex-test switch` is working properly
- Try `vtex-test deprecate` an innexistent app to check switching accounts (`vtex-test deprecate basedevmkp.asdf@0.0.1`
- Try `vtex-test undeprecate` an innexistent app to check switching accounts (`vtex-test undeprecate basedevmkp.asdf@0.0.1`
- Try `vtex-test publish` an app with typescript errors to check switching accounts

#### Types of changes

#### Chores checklist
- [X] Update `CHANGELOG.md`


src/lib/session/SessionManager.ts
```diff
@@ -94,29 +120,74 @@ export class SessionManager {
...
+      this.state.lastAccount = this.account
+      this.state.lastWorkspace = null
+    }
+
+    const cachedToken = new Token(this.sessionPersister.getAccountToken(newAccount))
+    if (useCachedToken && cachedToken.isValid()) {
+      this.state.account = newAccount
+      this.state.workspace = targetWorkspace
+      this.state.tokenObj = cachedToken
+      this.flushState()
```
Jeymisson, line 133: This name `flushState` gives the impression we are disposing the state, but we are actually refreshing it.

tiagonapoli: How about `saveState`? The idea of `flushState` was to send it to the underlying sessionPersister implementation to save it to disk, for example

Jeymisson: I think this one works better. `saveState` or `updateState`.

src/modules/support/login.ts
```diff
@@ -60,17 +62,18 @@ const assertToken = (raw: string): void => {
 }
 
 const saveSupportCredentials = (account: string, token: string): void => {
-  saveAccount(account)
-  saveWorkspace('master')
-  saveToken(token)
+  const session = SessionManager.getSingleton()
+  session.DEPRECATEDchangeAccount(account)
```
Jeymisson, line 66: Do you really want the deprecated one?

tiagonapoli: In this case I don't really know how `support login` works, nor how to test it - I created the deprecated one just for this. In a future PR we would want to change it

tiagonapoli: This should use `session.login`, providing a login method, implemented here:
https://github.com/vtex/toolbelt/blob/e3abd57ab55eb80d648f7f53d20e1c6054090b6a/src/lib/auth/AuthProviders/AuthProviders.ts#L5-L9

PR #854, opened 2020-05-07
test e2e Screenshot and video assets updated according to tester-hub

TRibeiral:

tester-hub was updated, changing the way the test results assets are accessed by the user. So the toolbelt `vtex test e2e` must be changed accordingly.

An app with cypress tests must be used, I recommend `shipping-preview`
add `"cypress": "0.x"` builder to its manifest
ensure that there is no   `"video": false` parameter on `cypress.json`
The app must be linked first then you should use the command `vtex tet e2e`

it should appear a working link to the video and screenshot assets at the test results

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [vtex test e2e] Show screenshot and videos assets URLs received from `vtex.tester-hub` new API.
 
 ## Fixed
-- Show screenshot and videos assets URLs received from `vtex.tester-hub` new API.
+- [vtex test e2e] Show screenshot and videos assets URLs received from `vtex.tester-hub` new API.
 
 ## [2.99.3] - 2020-05-05
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Fixed
 - [vtex test e2e] Show screenshot and videos assets URLs received from `vtex.tester-hub` new API.
 
-## Fixed
-- [vtex test e2e] Show screenshot and videos assets URLs received from `vtex.tester-hub` new API.
-
 ## [2.99.3] - 2020-05-05
 ### Fixed
 - [telemetry] Sanitize jwt tokens and simplify buffers serializtion.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Updated
 - [oclif] `oclif-plugin-spaced-commands` to fix alias documentation.
 
+## Fixed
+- [vtex test e2e] Screenshot and video assets updated according to tester-hub.
+
 ## [2.99.3] - 2020-05-05
 ### Fixed
 - [telemetry] Sanitize jwt tokens and simplify buffers serializtion.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [vtex test e2e] Show screenshot and videos assets URLs received from `vtex.tester-hub` new API.
 
 ## Fixed
-- [vtex test e2e] Screenshot and video assets updated according to tester-hub.
+- Show screenshot and videos assets URLs received from `vtex.tester-hub` new API.
 
 ## [2.99.3] - 2020-05-05
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Fixed
 - [vtex test e2e] Show screenshot and videos assets URLs received from `vtex.tester-hub` new API.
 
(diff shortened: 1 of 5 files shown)


src/modules/apps/e2e/report/failedApps.tsx
```diff
@@ -30,11 +31,14 @@ const FailedSpecDetail: React.FunctionComponent<SpecDetailProps> = ({ label, tex
 interface SpecProps {
   spec: string
   report: SpecReport
+  hubTestId: string
 }
 
-export const FailedSpec: React.FunctionComponent<SpecProps> = ({ spec, report }) => {
+export const FailedSpec: React.FunctionComponent<SpecProps> = ({ spec, report, hubTestId }) => {
+  const { logId } = report
+  const { specId } = report
```
tiagonapoli, line 39: _nit_:
```suggestion
  const { logId, specId } = report
```

src/modules/apps/e2e/report/failedApps.tsx
```diff
@@ -80,17 +90,28 @@ const ErrorVisualization: React.FunctionComponent<ErrorVisualizationProps> = ({
...
 }) => {
-  const testScreenshotsText = testScreenshots.map(curScreenshot => ` ${curScreenshot.path}`).join('\n')
+  const { account, workspace } = SessionManager.getSingleton()
+  const testScreenshotsUrl = testScreenshots
+    .map(
+      curScreenshot =>
+        `${workspace}--${account}.myvtex.com/_v/screenshot/${testId}/${specId}/${curScreenshot.screenshotId}`
+    )
+    .join('\n')
+  const videoUrl = testVideo === 'true' ? `${workspace}--${account}.myvtex.com/_v/video/${testId}/${specId}` : 'false'
```
tiagonapoli, line 104: testVideo is a string with 'false' or 'true'?

src/modules/apps/e2e/report/failedApps.tsx
```diff
@@ -80,17 +90,28 @@ const ErrorVisualization: React.FunctionComponent<ErrorVisualizationProps> = ({
...
     <Box key={title.join('')} flexDirection="column">
       <FailedSpecDetail label="Test" text={title.join(' ')} indented={false} />
       {body && <FailedSpecDetail label="Body" text={body} indented />}
       {stack && <FailedSpecDetail label="Stack" text={stack} indented />}
       {error && <FailedSpecDetail label="Error" text={error} indented />}
       {testScreenshots.length > 0 && (
-        <FailedSpecDetail label="Screenshots" text={testScreenshotsText} indented={testScreenshots.length > 1} />
+        <FailedSpecDetail label="Screenshots" text={testScreenshotsUrl} indented={testScreenshots.length > 1} />
       )}
+      {testVideo && <FailedSpecDetail label="Video" text={videoUrl} indented={false} />}
```
tiagonapoli, line 114: If testVideo is 'false' this will still create the video components

testVideo could be converted to boolean, or this could be `testVideo === 'true' && <FailedSpec...`

src/modules/apps/e2e/report/failedApps.tsx
tiagonapoli: I think videoUrl can be `''` if testVideo is not 'true'

src/modules/apps/e2e/report/failedApps.tsx
```diff
@@ -80,17 +90,28 @@ const ErrorVisualization: React.FunctionComponent<ErrorVisualizationProps> = ({
...
   testScreenshots,
+  testVideo,
+  specId,
+  testId,
 }) => {
-  const testScreenshotsText = testScreenshots.map(curScreenshot => ` ${curScreenshot.path}`).join('\n')
+  const { account, workspace } = SessionManager.getSingleton()
+  const testScreenshotsUrl = testScreenshots
+    .map(
+      curScreenshot =>
```
tiagonapoli, line 100: Could you create a function:
```
const workspaceBaseURL = () => {
  const { account, workspace } = SessionManager.getSingleton()
  return `${workspace}--${account}.${publicEndpoint()}`
}
```
Ideally this should be a common function, but for now there isn't one like that :(

(the `publicEndpoint` function is from `env.ts`)

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Fixed
+- [vtex test e2e] Screenshot and video assets updated according to tester-hub.
```
tiagonapoli, line 10: Didn't quite understand - how about something like:
```
Show screenshot and videos assets URLs received from `vtex.tester-hub` new API.
```
Is this correct?

src/modules/apps/e2e/report/failedApps.tsx
TRibeiral: Yes, testVideo is a string with 'false' or 'true'. 
I changed the videoUrl as suggested

CHANGELOG.md
TRibeiral: yes, changed it

src/modules/apps/e2e/report/failedApps.tsx
TRibeiral: 'true'

tiagonapoli: Here `testVideo` would have to be compared against 'true', because in case testVideo os 'false' the component would be rendered anyway

PR #856, opened 2020-05-08
Fix undeprecate 403. Add telemetry on deprecate and undeprecate

tiagonapoli:
The token used by `vtex deprecate` was the one from the previous account (when switching is necessary for deprecating an app) - this was fixed.

Also, telemetry was added on `vtex deprecate` and `vtex undeprecate` errors.

```
git checkout chore/telemetry-undeprecate

yarn watch
```
 
From an account different than `vtex`, run:
```
vtex-test deprecate vtex.service-example@0.1.0-beta.0
vtex-test undeprecate vtex.service-example@0.1.0-beta.0
```

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/src/modules/apps/undeprecate.ts
+++ b/src/modules/apps/undeprecate.ts
@@ -49,7 +49,7 @@ const prepareUndeprecate = async (appsList: string[]): Promise<void> => {
       log.info('Successfully undeprecated', app)
     } catch (e) {
       const errReport = ErrorReport.create({ originalError: e })
-      
+
       if (e.response && e.response.status && e.response.status === 404) {
         log.error(`Error undeprecating ${app}. App not found.`)
         errReport.logErrorForUser({ coreLogLevelDefault: 'debug' })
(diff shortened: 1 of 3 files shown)


src/modules/apps/undeprecate.ts
```diff
@@ -68,6 +68,8 @@ const prepareUndeprecate = async (appsList: string[]): Promise<void> => {
       }
     }
   }
+
+  await returnToPreviousAccount({ previousAccount: originalAccount, previousWorkspace: originalWorkspace })
```
rafaelbac, line 72: I noticed that we have a call to `returnToPreviousAccount()` at the end of the `undeprecate` command, but we do not have it for the `deprecate`. Is that on purpose?
With this difference, for the `deprecate` command we'll not return to the previous account either when the status is 404 or when it's successful, but we will do it for the `undeprecate`.

If I remember correctly, we want to return to the previous account independently of success or failure. Is that correct?
If that is the case, what do you think about adding a `finally` block to the `try/catch`? We could call the `returnToPreviousAccount()` there.

PR #868, opened 2020-05-22
Create workspace when logging in into a non-existent one

tiagonapoli:
- Create workspace if the targetWorkspace doesn't exist when using `vtex switch` or `vtex login` 
- Do not send `CommandError`s (usually misusage errors) to telemetry.

```
yarn global add vtex@2.101.1-beta
```
Test around the commands:
```
vtex login account -w nonExistentWorkspace
vtex login account -w existentWorkspace

vtex switch account -w nonExistentWorkspace
vtex switch account -w existentWorkspace

vtex workspace create nonExistentWorkspace
vtex workspace create existentWorkspace

vtex workspace use existentWorkspace
vtex workspace use nonExistentWorkspace
```

#### Types of changes

#### Chores checklist
- [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/lib/error/ErrorReport.ts
+++ b/src/lib/error/ErrorReport.ts
@@ -11,6 +11,15 @@ import logger from '../../logger'
 import { SessionManager } from '../session/SessionManager'
 import { getPlatform } from '../utils/getPlatform'
 import { TelemetryCollector } from '../telemetry/TelemetryCollector'
+import { CommandError } from '../../errors'
+
+interface CustomErrorReportCreateArgs extends ErrorReportCreateArgs {
+  shouldRemoteReport?: boolean
+}
+
+interface CustomErrorReportBaseConstructorArgs extends ErrorReportBaseConstructorArgs {
+  shouldRemoteReport: boolean
+}
 
 interface ErrorEnv {
   account: string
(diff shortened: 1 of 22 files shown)


src/modules/workspace/create.ts
```diff
@@ -16,31 +25,87 @@ const warmUpRouteMap = async (workspace: string) => {
   } catch (err) {} // eslint-disable-line no-empty
```
VerasThiago, line 25: It's ok to no handle this error ?

src/lib/error/ErrorReport.ts
```diff
@@ -39,15 +48,27 @@ interface LogToUserOptions {
 }
 
 export class ErrorReport extends ErrorReportBase {
-  public static create(args: ErrorReportCreateArgs) {
-    return new ErrorReport(createErrorReportBaseArgs(args))
+  public static checkIfShouldRemoteReport(err: Error | any) {
```
VerasThiago, line 51: why you chose 
```
if (err instanceof CommandError) {
  return false
}

return true
```
instead of 
`
return (err instanceof CommandError == false)
`

tiagonapoli: For now there's only `CommandError` type that shouldn't be reported, but we may want to add more types later - I think the more verbose way to check the instances is more understandable in this case

src/modules/workspace/create.ts
tiagonapoli: I think in this case it is - that wasn't changed from the original code. This code sends a request to the builder-hub on the recently changed workspace so Apps will warm up the route mapping (Apps has a route map for all workspaces - this is usually cached)

tiagonapoli: The user doesn't have to wait for this request to wait, so I think it's ok to not handle this error

PR #870, opened 2020-05-27
Session debug

VerasThiago:
Allow debugging development session on Toolbelt 

None

Link or relink any app and check the `buid.vtex-id` on `Jaeger`

#### Types of changes

#### Chores checklist
- [X] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [vtex login, vtex switch] Prompt workspace creation if it doesn't exist.
 
 ### Added
-- [vtex link] Generate `linkID` and set the `VTEX_LINK_ID` header request to `builder-hub`
-- [vtex re-link] Set the `VTEX_LINK_ID` header request to `builder-hub`
+- [vtex link] Generate `linkID` and set the `x-vtex-bh-link-id` header on link and relink requests to `builder-hub`.
 
 ## [2.101.0] - 2020-05-19
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - [vtex login, vtex switch] Prompt workspace creation if it doesn't exist.
 
+### Added
+- [vtex link] Generate `linkID` and set the `VTEX_LINK_ID` header request to `builder-hub`
+- [vtex re-link] Set the `VTEX_LINK_ID` header request to `builder-hub`
+
 ## [2.101.0] - 2020-05-19
 ### Added
 - [vtex publish] Message referencing location of the published docs if any.
--- a/src/lib/clients/IOClients/apps/Builder.ts
+++ b/src/lib/clients/IOClients/apps/Builder.ts
@@ -26,6 +26,14 @@ export interface BuildResult {
   timeNano?: number
 }
 
+export interface PrepareRequest {
+  zipFile?: Buffer
+  builderHubResolvingOpts: {
+    sticky: boolean
+    stickyHint: string
+  }
+}
+
 export interface AvailabilityResponse {
   host: string | undefined
   hostname: string | undefined
@@ -108,23 +116,18 @@ export class Builder extends AppClient {
 
   public linkApp = (
     app: string,
+    linkID: string,
     zipFile: Buffer,
     stickyOptions: StickyOptions = { sticky: true },
-    params: RequestParams = {},
-    requestHeaders: Record<string, any> = {}
-  ) => {
-    return this.sendZipFile(routes.link(app), app, zipFile, stickyOptions, params, requestHeaders)
-  }
-
-  public relinkApp = (
-    app: string,
-    changes: ChangeToSend[],
-    headersParams: Record<string, any>,
     params: RequestParams = {}
   ) => {
+    return this.sendZipFile(routes.link(app), app, zipFile, stickyOptions, params, { [Headers.VTEX_LINK_ID]: linkID })
+  }
+
+  public relinkApp = (app: string, changes: ChangeToSend[], linkID: string, params: RequestParams = {}) => {
     const headers = {
       ...(this.stickyHost && { [Headers.VTEX_STICKY_HOST]: this.stickyHost }),
-      ...headersParams,
+      ...{ [Headers.VTEX_LINK_ID]: linkID },
       'Content-Type': 'application/json',
     }
     const metric = 'bh-relink'
--- a/src/lib/clients/IOClients/apps/Builder.ts
+++ b/src/lib/clients/IOClients/apps/Builder.ts
@@ -14,24 +14,11 @@ export interface RequestParams {
   skipSemVerEnsure?: boolean
 }
 
-export interface BuilderHubRequestOptions {
-  params?: RequestParams
-  headers?: Record<string, any>
-}
-
 export interface BuildResult {
   availability?: AvailabilityResponse
   code?: string
   message?: any
   timeNano?: number
-}
-
-export interface PrepareRequest {
-  zipFile?: Buffer
-  builderHubResolvingOpts: {
-    sticky: boolean
-    stickyHint: string
-  }
 }
 
 export interface AvailabilityResponse {
--- a/src/lib/clients/IOClients/apps/Builder.ts
+++ b/src/lib/clients/IOClients/apps/Builder.ts
@@ -1,4 +1,4 @@
-import { randomBytes } from "crypto";
+import { randomBytes } from 'crypto'
 import { AppClient, CacheType, InstanceOptions, IOContext } from '@vtex/api'
 import { ChangeToSend } from '../../../../modules/apps/ProjectUploader'
 import { Headers } from '../../../constants/Headers'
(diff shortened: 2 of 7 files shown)


src/lib/clients/IOClients/apps/Builder.ts
```diff
@@ -111,9 +114,11 @@ export class Builder extends AppClient {
   }
 
   public relinkApp = (app: string, changes: ChangeToSend[], params: RequestParams = {}) => {
+    console.info(`linkID: ${this.linkID}`)
```
tiagonapoli, line 117: The client shouldn't be responsible for logging

src/lib/clients/IOClients/apps/Builder.ts
```diff
@@ -47,9 +48,11 @@ export class Builder extends AppClient {
   }
 
   private stickyHost!: string
+  private linkID: string
```
tiagonapoli, line 51: I think a linkId should be held not by the client, but maybe `ProjectUploader`

src/lib/clients/IOClients/apps/Builder.ts
```diff
@@ -47,9 +48,11 @@ export class Builder extends AppClient {
...
 
   constructor(ioContext: IOContext, opts?: InstanceOptions) {
     super('vtex.builder-hub@0.x', ioContext, opts)
+    this.linkID = randomBytes(20).toString('hex')
```
tiagonapoli, line 55: I think 20 bytes for this is maybe too much - 20 bytes = 160 bits = 10^48. I think 32, or 64 bits, is enough, so 4bytes or 8 bytes

src/lib/clients/IOClients/apps/Builder.ts
tiagonapoli: I think the caller should be responsible

tiagonapoli: Also, when logging on toolbelt, use the `logger` object - there's a `logger.ts` that exports a logger

src/lib/clients/IOClients/apps/Builder.ts
tiagonapoli: Maybe the client should receive a linkID parameter on `relinkApp` and `linkApp`

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- [vtex link] Generate `linkID` and set the `VTEX_LINK_ID` header request to `builder-hub`
+- [vtex re-link] Set the `VTEX_LINK_ID` header request to `builder-hub`
```
tiagonapoli, line 11: I think one entry is ok:
```suggestion
- [vtex link] Generate `linkID` and set the `x-vtex-bh-link-id` header on link and relink requests to `builder-hub`.
```

src/lib/constants/Headers.ts
```diff
@@ -10,4 +10,7 @@ export enum Headers {
   // Specify the target cluster for the request.
   // Works only on myvtexdev domain
   VTEX_UPSTREAM_TARGET = 'x-vtex-upstream-target',
+
+  // ID of link to keep on track all re-links related to that link.
```
tiagonapoli, line 14: How about something like:
```
// LinkID used by builder-hub to create logs/tracing correlation between all relinks related to a initial link
```

src/lib/clients/IOClients/apps/Builder.ts
```diff
@@ -105,15 +110,22 @@ export class Builder extends AppClient {
...
   ) => {
-    return this.sendZipFile(routes.link(app), app, zipFile, stickyOptions, params)
+    return this.sendZipFile(routes.link(app), app, zipFile, stickyOptions, params, requestHeaders)
   }
 
-  public relinkApp = (app: string, changes: ChangeToSend[], params: RequestParams = {}) => {
+  public relinkApp = (
+    app: string,
+    changes: ChangeToSend[],
+    headersParams: Record<string, any>,
```
tiagonapoli, line 122: `requestHeaders`, or just `headers`

tiagonapoli: I think `relinkApp` and `linkApp`, instead of having `headers` as parameter they could have `linkId` as parameter, and then call:
```ts
return this.sendZipFile(routes.link(app), app, zipFile, stickyOptions, params, { [Headers.VTEX_LINK_ID]: linkID }
```
The difference in these approachs is semantic:
- Using headers: It seems a optional thing - the function user can pass any header it wants, not specifically the link-id header
- Using linkId as parameter: The function user *has* to provide a linkId on this operation - which is what we want to enforce

src/lib/constants/Headers.ts
```diff
@@ -10,4 +10,7 @@ export enum Headers {
   // Specify the target cluster for the request.
   // Works only on myvtexdev domain
   VTEX_UPSTREAM_TARGET = 'x-vtex-upstream-target',
+
+  // LinkID used by builder-hub to create logs/tracing correlation between all relinks related to a initial link
```
tiagonapoli, line 14:
```suggestion
  // LinkID used by builder-hub to create logs/tracing correlation
  // between all relink/link operations related to a initial link
```

src/modules/apps/ProjectUploader.ts
```diff
@@ -72,21 +72,30 @@ export class ProjectUploader {
...
   }
 
-  public sendToLink(files: FileToSend[], requestParams: RequestParams = {}) {
-    return this.sendWholeProject('link', files, requestParams)
+  public sendToLink(files: FileToSend[], builderHubRequestOptions: BuilderHubRequestOptions = {}) {
+    return this.sendWholeProject('link', files, builderHubRequestOptions)
   }
 
-  public sendToRelink(changes: ChangeToSend[], requestParams: RequestParams = {}) {
+  public sendToRelink(changes: ChangeToSend[], builderHubRequestOptions: BuilderHubRequestOptions = {}) {
```
tiagonapoli, line 91: I think this could be:
```ts
public sendToRelink(changes: ChangeToSend[], linkId: string, requestParams: RequestParams = {})
```

src/modules/apps/link.ts
```diff
@@ -81,7 +84,11 @@ const performInitialLink = async (
     }
 
     try {
-      const { code } = await projectUploader.sendToLink(filesWithContent, { tsErrorsAsWarnings: unsafe })
+      console.info(`link ID: ${linkID}`)
```
tiagonapoli, line 87: Use `log.info`

src/modules/apps/link.ts
```diff
@@ -163,8 +170,10 @@ const watchAndSendChanges = async (
 
   const sendChanges = debounce(async () => {
     try {
+      console.info(`link ID: ${linkID}`)
```
tiagonapoli, line 173: Use `log.info`

src/modules/apps/link.ts
```diff
@@ -1,27 +1,29 @@
...
+import { randomBytes } from 'crypto'
+import { readFileSync } from 'fs'
+import { YarnFilesManager } from '../../lib/files/YarnFilesManager'
+import chalk from 'chalk'
+import chokidar from 'chokidar'
+import debounce from 'debounce'
+import log from '../../logger'
+import moment from 'moment'
+import retry from 'async-retry'
 import startDebuggerTunnel from './debugger'
```
tiagonapoli, line 26: Why all import orders was changed?

src/lib/clients/IOClients/apps/Builder.ts
```diff
@@ -14,13 +14,26 @@ export interface RequestParams {
   skipSemVerEnsure?: boolean
 }
 
+export interface BuilderHubRequestOptions {
```
tiagonapoli, line 17: Is this interface still in use?

src/lib/clients/IOClients/apps/Builder.ts
```diff
@@ -14,13 +14,26 @@ export interface RequestParams {
...
+}
+
 export interface BuildResult {
   availability?: AvailabilityResponse
   code?: string
   message?: any
   timeNano?: number
 }
 
+export interface PrepareRequest {
```
tiagonapoli, line 29: This is used only on projectUploader, right? You can move it there, don't need to export it

src/modules/apps/link.ts
```diff
@@ -163,7 +166,8 @@ const watchAndSendChanges = async (
 
   const sendChanges = debounce(async () => {
     try {
-      return await projectUploader.sendToRelink(changeQueue.splice(0, changeQueue.length), {
+      log.info(`link ID: ${linkID}`)
```
tiagonapoli, line 169: nit:
```suggestion
      log.info(`Link ID: ${linkID}`)
```

src/modules/apps/link.ts
```diff
@@ -81,7 +83,8 @@ const performInitialLink = async (
     }
 
     try {
-      const { code } = await projectUploader.sendToLink(filesWithContent, { tsErrorsAsWarnings: unsafe })
+      log.info(`link ID: ${linkID}`)
```
tiagonapoli, line 86: nit:
```suggestion
      log.info(`Link ID: ${linkID}`)
```

src/lib/clients/IOClients/apps/Builder.ts
VerasThiago: Not anymore, I will remove

src/lib/clients/IOClients/apps/Builder.ts
VerasThiago: Sure!

src/lib/clients/IOClients/apps/Builder.ts
```diff
@@ -14,13 +14,26 @@ export interface RequestParams {
...
+  zipFile?: Buffer
```
tiagonapoli, line 30: Is zipFile optional?

VerasThiago: No, I saw that and forgot to change, sorry!

PR #874, opened 2020-06-02
Feat/trace flag

VerasThiago:
Creating a new flag `--trace` that set the `jaeger-debug-id`

none

#### Types of changes

#### Chores checklist
- [X] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- [CustomCommand] new flag `trace` that set the `jaeger-debug-id` header automatically
+- [All Commands] new flag `--trace` that sets the `jaeger-debug-id` header automatically on all requests.
 
 ## [2.101.1] - 2020-05-28
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- [CustomCommand] new flag `trace` that set the `jaeger-debug-id` header automatically
+
 ## [2.101.1] - 2020-05-28
 ### Added
 - [vtex test e2e] Added retry count to account for cold `tester-hub` start.
--- a/src/lib/constants/Headers.ts
+++ b/src/lib/constants/Headers.ts
@@ -14,4 +14,7 @@ export enum Headers {
   // LinkID used by builder-hub to create logs/tracing correlation between all
   // relink/link operations related to a initial link
   VTEX_LINK_ID = 'x-vtex-bh-link-id',
+
+  // Tracing header to be used by jaeger
+  VTEX_TRACE = 'jaeger-debug-id'
 }
--- a/src/lib/constants/Headers.ts
+++ b/src/lib/constants/Headers.ts
@@ -16,5 +16,5 @@ export enum Headers {
   VTEX_LINK_ID = 'x-vtex-bh-link-id',
 
   // Tracing header to be used by jaeger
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- [CustomCommand] new flag `trace` that set the `jaeger-debug-id` header automatically
```
tiagonapoli, line 10: I'm thinking that maybe our users that read the changelog wouldn't understand the `CustomCommand` tag, maybe something like:
```suggestion
- [All Commands] new flag `--trace` that sets the `jaeger-debug-id` header automatically on all requests.
```

src/lib/constants/Headers.ts
```diff
@@ -14,4 +14,7 @@ export enum Headers {
   // LinkID used by builder-hub to create logs/tracing correlation between all
   // relink/link operations related to a initial link
   VTEX_LINK_ID = 'x-vtex-bh-link-id',
+
+  // Tracing header to be used by jaeger
```
tiagonapoli, line 18:
```suggestion
  // Header to ensure that the request trace will be sampled
```

src/lib/constants/Headers.ts
```diff
@@ -14,4 +14,7 @@ export enum Headers {
...
+  VTEX_TRACE = 'jaeger-debug-id'
```
tiagonapoli, line 19: Since the header ensures that the trace will be sampled (recorded on our backend), how about:
```ts
VTEX_ENSURE_TRACE = 'jaeger-debug-id'
```

src/lib/globalConfigs/traceConfig.ts
```diff
@@ -0,0 +1,24 @@
+import { randomBytes } from 'crypto'
+import log from './../../logger'
+
+export class TraceConfig {
+    public static jeagerDebugID: string
```
tiagonapoli, line 5:
```suggestion
    public static jaegerDebugID: string
```

src/oclif/CustomCommand.ts
```diff
@@ -1,25 +1,33 @@
...
+import { TelemetryCollector } from '../lib/telemetry/TelemetryCollector'
+import { TraceConfig } from '../lib/globalConfigs/traceConfig';
+import * as Parser from '@oclif/parser';
+import OclifCommand, { flags as oclifFlags } from '@oclif/command'
 
 export abstract class CustomCommand extends OclifCommand {
   public static globalFlags = {
     verbose: oclifFlags.boolean({ char: 'v', description: 'Show debug level logs', default: false }),
     help: oclifFlags.help({ char: 'h' }),
+    trace: oclifFlags.boolean({ char: 't', description: 'Log tracing to jaeger', default: false }),
```
tiagonapoli, line 14: I think the trace flag doesn't need a short version - this would remove the `-t` flag for all commands

tiagonapoli, line 14: We don't have to expose to common users that we use jaeger :) (a advanced user can easily find that info, though :cry: )
```suggestion
    trace: oclifFlags.boolean({ char: 't', description: 'Ensure all HTTP requests to VTEX IO are traced', default: false }),
```

src/lib/globalConfigs/traceConfig.ts
```diff
@@ -0,0 +1,23 @@
...
+  public static jeagerDebugID: string
+  public static traceFlag = false
+
+  public static setShouldTrace() {
+    this.traceFlag = true
+    this.jeagerDebugID = `toolbelt-${randomBytes(8).toString('hex')}`
+    log.info(`Trace Debug ID: ${this.jeagerDebugID}`)
+  }
+
+  public static checkTrace(traceFlag: boolean) {
```
tiagonapoli, line 14: How about:
```ts
public static setupTraceConfig(
```

PR #875, opened 2020-06-04
Handle error on unlinked apps when running `vtex test e2e`

wdsrocha:

Raise the proper error when running `vtex test e2e` on an app that isn't linked. Without this fix, it might be not so clear why the script is breaking. Motivation for this PR: https://github.com/vtex-apps/checkout-io-tests/pull/5#issuecomment-638850876

Before:

image

After:

Screen Shot 2020-06-04 at 11 21 30

1. Clone vtex-apps/checkout-io-tests#5
2. Go to `checkoutio` account on `wdsrocha` workspace
3. Run `yarn e2e:run:remote`

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/src/modules/apps/e2e/index.tsx
+++ b/src/modules/apps/e2e/index.tsx
@@ -29,7 +29,7 @@ class EndToEndCommand {
     const { data: workspaceAppsList } = await apps.listApps()
     const appItem = workspaceAppsList.find(({ app }) => app.startsWith(cleanAppId))
 
-    if (appItem.id === undefined) {
+    if (appItem === undefined) {
       throw new Error(`App "${cleanAppId}" was not found in the current workspace!`)
     }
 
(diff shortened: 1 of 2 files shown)


src/modules/apps/e2e/index.tsx
```diff
@@ -29,7 +29,7 @@ class EndToEndCommand {
     const { data: workspaceAppsList } = await apps.listApps()
     const appItem = workspaceAppsList.find(({ app }) => app.startsWith(cleanAppId))
 
-    if (appItem.id === undefined) {
+    if (appItem?.id === undefined) {
```
cezarguimaraes, line 32: _nit (optional)_: Could you change it to
```ts
if (appItem === undefined) {
```
It avoids the extra conditional and won't lead others (like myself just now 😂) to think that `id` might be undefined - I checked and it's always the same as `appItem.app`.

wdsrocha: Cool, I just assumed that `id` could be undefined. Will update!

PR #892, opened 2020-06-30
Feat/plugins | Breaking toolbelt commands into plugins

VerasThiago:
All documented [here](https://github.com/vtex/toolbelt/blob/feat/Plugins/docs/enhancements/0000-break-commands-into-plugins.md)
#### Types of changes

#### Chores checklist
- [X] Update `CHANGELOG.md`


docs/enhancements/0000-break-commands-into-plugins.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Motivation
+
+VTEX CLI is the first contact that we have with the VTEX IO products, so we need to have a clean and simple CLI that works pretty well.
+
+Nowadays we have a CLI that, by default, comes with a **LOT** of commands that probably most of them will never be used by the same user.
+
+![](https://i.imgur.com/2GIq2Ep.png)
+
+The idea is to remove this commands that are not necessary to come by default, i.e., `vtex lighthouse` and `vtex test`. Eventually the user will need to use some of this extra commands, so just import to your own CLI.
```
tiagonapoli, line 13:
```suggestion
The idea is to remove commands that are not necessary to be included on toolbelt by default, e.g., `vtex lighthouse` and `vtex test`. If the user eventually needs to use some of these extra commands, he/she can just add the desired command to its local CLI.
```

docs/enhancements/0000-break-commands-into-plugins.md
```diff
@@ -0,0 +1,132 @@
...
+
+We can also have a library with all the possible commands that the user can import such as [Adobe I/O CLI](https://github.com/adobe/aio-cli).
```
tiagonapoli, line 15:
```suggestion
In order to provide discoverability of possible commands the user can import a library can be created with all plugins approved by VTEX (inspired in  [Adobe I/O CLI](https://github.com/adobe/aio-cli)), this way the users will have a curated list of safe commands to install locally.
```

src/api/clients/IOClients/apps/index.ts
```diff
@@ -0,0 +1,8 @@
+export * from './ABTester'
+export * from './Billing'
+export * from './Builder'
+export * from './Rewriter'
+export * from './Sponsor'
+export * from './Tester'
```
tiagonapoli, line 6: This is specific to how the test plugin works, can be moved to there

src/api/clients/IOClients/apps/index.ts
```diff
@@ -0,0 +1,8 @@
...
+export * from './ToolbeltConfig'
+export * from './ToolbeltTelemetry'
```
tiagonapoli, line 8: Plugins will not need those two clients, they are toolbelt internals

src/api/manifest/ManifestUtil.ts
```diff
@@ -3,7 +3,7 @@ import { readFile, writeFile } from 'fs-extra'
 import path from 'path'
 import { memoize } from 'ramda'
 
-import { CommandError } from './errors'
+import { CommandError } from '../error/errors'
 
 const readFileUtf = (file: string): Promise<string> => {
```
tiagonapoli, line 8: These manifest utils are old, they should be replaced by the ManifestEditor class (also, the ManifestEditor class should be renamed to just Manifest)

tiagonapoli: So there's no need for these old manifest utils to be exported on the api

src/api/modules/utils.ts
```diff
@@ -244,3 +250,122 @@ export const resolveAppId = (appName: string, appVersion: string): Promise<strin
 }
 
 export const isLinked = propSatisfies<string, Manifest>(contains('+build'), 'version')
+
+// PS: All functions below was merged from the other utils.ts file
```
tiagonapoli, line 254: No need for this comment on the file (it could be a comment on the github PR though)

toolbelt
```diff
@@ -0,0 +1 @@
+/home/thiago/Documents/Git/toolbelt
```
tiagonapoli, line 1: this file can be removed

src/api/telemetry/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './TelemetryCollector'
+export * from './TelemetryStore'
+export * from './TelemetryReporter'
```
tiagonapoli, line 3: These two are toolbelt internals

src/api/telemetry/TelemetryReporter/report.ts
```diff
@@ -1,11 +1,11 @@
-const initTime = process.hrtime()
-
 import { TelemetryMetaMetrics } from '../../metrics/MetricNames'
 import { SessionManager } from '../../session/SessionManager'
 import { TelemetryLocalStore } from '../TelemetryStore'
 import { PendingTelemetryDataManager } from './PendingTelemetryDataManager'
 import { TelemetryReporter } from './TelemetryReporter'
 
+const initTime = process.hrtime()
```
tiagonapoli, line 7: This has to be placed on the beginning of the file

tiagonapoli: There are some rules overrides on .eslintrc that you may want to update the file locations:
https://github.com/vtex/toolbelt/blob/c0897a9056d01e98618b12d3bd2d450397223524/.eslintrc#L33-L47

src/api/telemetry/TelemetryReporter/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './PendingTelemetryDataManager'
```
tiagonapoli, line 1: this is internal

src/api/session/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './SessionManager'
+export * from './SessionsPersister'
```
tiagonapoli, line 2: toolbelt internal, no need to export

scripts/symlinkPlugin.sh
```diff
@@ -0,0 +1 @@
+ln -s $PWD $PWD/node_modules/vtex
```
tiagonapoli, line 1: Remember to try to delete the symlink before creating a new one

docs/enhancements/0000-break-commands-into-plugins.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Explanation
+
+This new change will be made by `plugins`, breaking all the commands and transform them into plugins, now we are going to add only necessary `plugins` as default. Some documentation about plugins was made by a study and is registered [here](https://github.com/VerasThiago/CLIBenchmark/blob/master/Oclif/docs/ABOUTME.md).
+
+## Internal details
+
+### Toolbelt Plugins
+
+Huge amount of toolbelt commands share the same piece of code, in order that, toolbelt plugins will need to use it as well. So what do toolbelt-plugins will need?
```
tiagonapoli, line 29:
```suggestion

Toolbelt commands share a lot of common code, which is a situation that will happen too when the commands are broken into plugins. Because of this we have to create a way to share this common code across plugins, but first we have to determine what we want to share across them:
```

docs/enhancements/0000-break-commands-into-plugins.md
```diff
@@ -0,0 +1,132 @@
...
+- `Namespaced Paths (?)` (create a plugins/plugin-name folder to place the folders the plugin want to use)
+- `Manifest reading/editing helpers` (ManifestEditor class, which ideally should be renamed to Manifest)
+- `PackageJson reading/editing helper`
+
+#### Approach #1
+Create a `@vtex/toolbelt-api` with all these needs. The plugins would always use this package and the toolbelt core would use this package as well.
+
+##### Problem #1
+
+How do we guarantee that all installed plugins are using the same `@vtex/toolbelt-api version`? The problem on them using different versions is the amount of code to be compiled when requiring and the fact that we could have fixed a major bug in a `@vtex/toolbelt-api`'s patch.
```
tiagonapoli, line 48:
```suggestion
How do we guarantee that all installed plugins are using the same `@vtex/toolbelt-api` version? The problem on them using different versions is the amount of code to be compiled when requiring and the fact that we could have fixed a major bug in a `@vtex/toolbelt-api`'s patch. 
```

docs/enhancements/0000-break-commands-into-plugins.md
```diff
@@ -0,0 +1,132 @@
...
+Solution with default plugins [here](https://github.com/VerasThiago/npmPackageTests/pull/1).
+
+Problem with external plugins [here](https://github.com/VerasThiago/npmPackageTests/pull/1#issuecomment-647696211)
+
+
+#### Approach #2 (**Chosen**)
+
+Reorganize toolbelt `/src`, to store the code that will be shared inside a folder `toolbelt-api` for example.
+
+Plugins now imports from `toolbelt` the code that they will use. Today the toolbelt don't allow this easily because he doesn't offer re-export entrypoints and don't have on the published bundle the typescript types.
```
tiagonapoli, line 85:
```suggestion
Plugins now import from `toolbelt` the code that they will use. Currently toolbelt doesn't allow this easily because it doesn't offer re-export entrypoints and doesn't have on the published bundle the typescript types.
```

docs/enhancements/0000-break-commands-into-plugins.md
```diff
@@ -0,0 +1,132 @@
...
+|:---|---:|---:|---:|---:|
+| `node simple-require.js` | 705.3 ± 31.2 | 662.0 | 818.0 | 1.69 ± 0.09 |
+| `node require-file.js` | 416.7 ± 13.6 | 389.3 | 450.4 | 1.00 |
+| `node destructure-require.js` | 696.5 ± 15.1 | 667.9 | 740.5 | 1.67 ± 0.07 |
+
+Check the full benchmark [here](https://github.com/tiagonapoli/benchmarking/tree/master/node-vtex-api-import
+)
+
+
+In order that, I made a benchmark to analyze if this cost will be really bad for the `toolbelt` performance, and the results was pretty good as far as I understand.
```
tiagonapoli, line 116:
```suggestion
Because of this, I made a benchmark to analyze if this cost will be really bad for the `toolbelt` performance, and the results was pretty good as far as I understand.
```

src/api/clients/IOClients/apps/index.ts
VerasThiago: Just remove this export will be enough ? Bexause they will be located at `api`folder. Do you think that creating the same folder inside `src/lib` it's a better approach ?

src/api/manifest/ManifestUtil.ts
VerasThiago: OK, just removed from the export

docs/enhancements/0000-break-commands-into-plugins.md
```diff
@@ -0,0 +1,132 @@
...
+
+VTEX CLI is the first contact that we have with the VTEX IO products, so we need to have a clean and simple CLI that works pretty well.
+
+Nowadays we have a CLI that, by default, comes with a **LOT** of commands that probably most of them will never be used by the same user.
+
+![](https://i.imgur.com/2GIq2Ep.png)
+
+The idea is to remove commands that are not necessary to be included on toolbelt by default, e.g., `vtex lighthouse` and `vtex test`. If the user eventually needs to use some of these extra commands, he/she can just add the desired command to its local CLI.
+
+In order to provide discoverability of possible commands the user can import a library can be created with all commands approved by VTEX (inspired in  [Adobe I/O CLI](https://github.com/adobe/aio-cli)), this way the users will have a curated list of safe commands to install locally.
```
tiagonapoli, line 15: Didn't quite understand this part:
```
the user can import a library can be created with all commands
```

PR #898, opened 2020-07-14
Fix settings set

tiagonapoli:
A lot of releases ago this command was changed, removing the value cast before setting it to a field. This PR reverts it, since the desired behavior is to cast the value before setting.

```
 vtex-test settings set vtex.service-example@0.x boolean false
```
You should see:
```json
{
  "fieldName": "fieldValue",
  "teste": 1,
  "boolean": false
}
```
Instead of:
```
{
  "fieldName": "fieldValue",
  "teste": 1,
  "boolean": "false"
}
```

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Fixed
 - [oclif:entrypoint] Only run oclif command when the entrypoint is the main script.
+- [vtex settings set] Cast value before setting field.
 
 ### Changed
 - [vtex edition] Refactor command to use as oclif plugin.
--- a/src/modules/apps/settings/set.ts
+++ b/src/modules/apps/settings/set.ts
@@ -1,11 +1,23 @@
 import { merge, __ } from 'ramda'
 import { createAppsClient } from '../../../api/clients/IOClients/infra/Apps'
 
+const castValue = (value: string) => {
+  let parsedValue: any
+  try {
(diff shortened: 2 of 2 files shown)


src/modules/apps/settings/set.ts
```diff
@@ -1,11 +1,23 @@
 import { merge, __ } from 'ramda'
 import { createAppsClient } from '../../../api/clients/IOClients/infra/Apps'
 
+const castValue = (value: string) => {
+  let parsedValue: any
+  try {
+    parsedValue = JSON.parse(value)
+  } catch (err) {
```
VerasThiago, line 8: You catch the err and make no moves with this, it's ok ?

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.105.1-beta.1] - 2020-07-14
```
VerasThiago, line 9: Don't forget to remove this releases

src/modules/apps/settings/set.ts
tiagonapoli: In this case it's ok, we try to parse a json - if we have an error that means that value isn't a json (check here the exceptions it throws https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse)

VerasThiago: Awesome!

PR #915, opened 2020-07-23
Feat/log flow error splunk

VerasThiago:
Logging the old `commandError` type to splunk to help identify flow issues in `toolbelt`

#### Types of changes

#### Chores checklist
- [ ] Update `CHANGELOG.md`


src/api/error/ErrorReport.ts
```diff
@@ -48,17 +48,14 @@ interface LogToUserOptions {
...
 export class ErrorReport extends ErrorReportBase {
-  public static checkIfShouldRemoteReport(err: Error | any) {
-    if (err instanceof CommandError) {
-      return false
-    }
-
-    return true
+  public static isFlowIssue(err: any) {
+    if (err instanceof ErrorReport && err.kind === ErrorKinds.FLOW_ISSUE_ERROR) return true
+    return false
```
tiagonapoli, line 53: Here I think we can go for the competitive programming way:
```ts
return err instanceof ErrorReport && err.kind === ErrorKinds.FLOW_ISSUE_ERROR
```

src/api/manifest/ManifestEditor.ts
```diff
@@ -28,15 +29,17 @@ export class ManifestEditor {
     }
   }
 
-  public static readAndParseManifest(path: string) {
+  public static async readAndParseManifest(path: string) {
     try {
-      return readJson(path)
+      return await readJson(path)
```
tiagonapoli, line 34: Wow, I think I was the responsible for this bug haha - I really learned a lot in the past couple months ,thanks for solving this

src/api/modules/utils.ts
```diff
@@ -75,15 +79,21 @@ export const validateAppAction = async (operation: string, app?): Promise<boolea
...
   }
 
   // No app arguments and no manifest file.
   const isReadable = await ManifestEditor.isManifestReadable()
   if (!app && !isReadable) {
-    throw new CommandError(
-      `No app was found, please fix your manifest.json${app ? ' or use <vendor>.<name>[@<version>]' : ''}`
-    )
+    throw ErrorReport.createAndMaybeRegisterOnTelemetry({
+      kind: ErrorKinds.FLOW_ISSUE_ERROR,
```
tiagonapoli, line 92: Maybe FlowIssues are so common that we would want a utility for create them?
You could create in error/utils.ts something like:
```ts
/** Create and register on telemetry an ErrorReport with errorKind = ErrorKinds.FLOW_ISSUE_ERROR */
export const createFlowIssueError(msg: string) {
  return ...
}
```

src/oclif/hooks/init.ts
```diff
@@ -148,31 +149,26 @@ export const onError = async (e: any) => {
...
-      default:
-        log.error('Unhandled exception')
-        log.error('Please report the issue in https://github.com/vtex/toolbelt/issues')
-        log.error('Raw error: ', e)
+  } else if (ErrorReport.isFlowIssue(e)) {
+    if (e.message && e.message !== '') {
+      log.error(e.message)
     }
+  } else if (e instanceof SSEConnectionError) {
+    log.error(e.message ?? 'Connection to login server has failed')
```
tiagonapoli, line 157: With the new login flow released in the last version this fallback is not necessary anymore, it can be just:
```ts
log.error(e.message)
```

tiagonapoli: Toolbelt doesn't communicate with the login server via sse anymore

src/api/error/ErrorKinds.ts
```diff
@@ -15,4 +15,5 @@ export const ErrorKinds = {
   APP_LOGS_SSE_ERROR: 'LogsSSEError',
   APP_LOGS_PARSE_ERROR: 'LogsParseError',
   STICKY_HOST_ERROR: 'StickyHostError',
+  FLOW_ISSUE_ERROR: 'FlowIssue',
```
lariciamota, line 18: What do you think about maintaining the pattern of the other constants? 
```suggestion
  FLOW_ISSUE_ERROR: 'FlowIssueError',
```

src/modules/workspace/create.ts
```diff
@@ -1,9 +1,9 @@
 import { Workspaces } from '@vtex/api'
 import chalk from 'chalk'
-import { CommandError } from '../../api/error/errors'
+import { createFlowIssueError } from '../../api/error'
+import { ErrorReport } from '../../api/error/ErrorReport'
```
lariciamota, line 4: Would it be possible to import like:
```suggestion
import { createFlowIssueError, ErrorReport } from '../../api/error'
```
?

src/api/manifest/ManifestEditor.ts
```diff
@@ -1,6 +1,6 @@
 import { readJson, writeJson, writeJsonSync } from 'fs-extra'
 import { resolve } from 'path'
-import { CommandError } from '../error/errors'
+import { createFlowIssueError } from '../error/utils'
```
lariciamota, line 3: Why in some cases this is being imported from ../error and in this case utils is being specified?

src/oclif/hooks/init.ts
```diff
@@ -148,31 +149,26 @@ export const onError = async (e: any) => {
...
-        break
-      case SSEConnectionError.name:
-        log.error(e.message ?? 'Connection to login server has failed')
-        break
-      default:
-        log.error('Unhandled exception')
-        log.error('Please report the issue in https://github.com/vtex/toolbelt/issues')
-        log.error('Raw error: ', e)
+  } else if (ErrorReport.isFlowIssue(e)) {
+    if (e.message && e.message !== '') {
```
lariciamota, line 153: Is this checking `e.message !== ''` necessary?

src/api/error/ErrorKinds.ts
VerasThiago: I think that this a good Idea, but we discussed this already with Molgado and they decided to use without `Error`

src/oclif/hooks/init.ts
VerasThiago: This was necessary in the old toolbelt login flow, but with the new one I don't know exactly if it's necessary or not. Napoli can say a little bit more about it

src/modules/workspace/create.ts
VerasThiago: Yes it is. I used to import `ErrorReport` and `ErrorKind` as `import {ErrorReport, ErrorKind} from '../../api/error` but this can cause a ciclyc  import, so to solve this problem the correct way is to import from separeted paths, so I'm going to fix this to the full path

src/api/manifest/ManifestEditor.ts
VerasThiago: Sorry, going to fix this

PR #922, opened 2020-07-30
Add docs for toolbelt maintainers

tiagonapoli:
Recipes for:
- Deprecating a version
- Enforcing a new version


docs/MAINTAINING.md
```diff
@@ -0,0 +1,76 @@
+## Maintaining VTEX Toolbelt
+
+This document is for people working on VTEX as Toolbelt maintainers.
+
+### Forcing a new version
+
+Toolbelt has a way of enforcing users to use a minimum version choosed by us. This enforcement is done on the client side (can be disabled by an environment variable, check [EnvVariables](../src/lib/constants/EnvVariables.ts)) and on `vtex.builder-hub` - users using an older toolbelt version will not be allowed to link or publish. This feature is useful for enforcing users to update to the latest version, which can be important when we launch a new version with a major bug fix for example.
```
VictoriaMolgado, line 7: Maybe we should explain why we do that? As someone with less context, I read that and now I'm thinking: ok, we make users have a minimum version, but why is it good for them? Should we continue doing that?

VictoriaMolgado, line 7: I understood why it is interesting for us, but why is it interesting for users? Maybe: 

"This feature is useful for enforcing users to update to the latest version, which can be important when we launch a new version with a major bug fix for example, making then know quickly we are improving their experiences, saving time to read documentations about new releases." ?

docs/MAINTAINING.md
```diff
@@ -0,0 +1,76 @@
...
+This version 2.93.0 was deprecated. Please update to the latest version: 'yarn global add vtex'.
+```
+
+This enforcement is on the client side and can be disabled on a execution by setting an environment variable (check check [EnvVariables](../src/lib/constants/EnvVariables.ts)).
+
+There's a problem with this approach - when we deprecate a npm package version, the `latest` tag is not automatically relocated, we have to do it manually, so the full process of deprecating a version is the following (the version chosen is just an example):
+
+1. Deprecate it:
+```
+npm deprecate vtex@2.92.0 -m "Message explaining why this version was deprecated, for example: Bug when sending files to link"
```
VerasThiago, line 66: I tested in a test package and got an error, I thing the correct way is without `-m`
npm deprecate vtex@2.92.0 "message here"

docs/MAINTAINING.md
```diff
@@ -0,0 +1,76 @@
+## Maintaining VTEX Toolbelt
+
+This document is for people working on VTEX as Toolbelt maintainers.
+
+### Forcing a new version
+
+Toolbelt has a way of enforcing users to use a minimum version choosed by us. This enforcement is done on the client side (can be disabled by an environment variable, check [EnvVariables](../src/lib/constants/EnvVariables.ts)) and on `vtex.builder-hub` - users using an older toolbelt version will not be allowed to link or publish. This feature is useful for enforcing users to update to the latest version, which can be important when we launch a new version with a major bug fix for example.
+
+Here's how you enforce minumum toolbelt version:
```
lariciamota, line 9: Just a typo
```suggestion
Here's how you enforce minimum toolbelt version:
```

docs/MAINTAINING.md
```diff
@@ -0,0 +1,76 @@
...
+$ toolbelt-conf config:get
+
+All configs
+{
+  minimumToolbeltVersion: '2.102.1',
+  minimumToolbeltPrereleaseVersion: '2.102.1'
+}
+```
+
+The `minimumToolbeltPrereleaseVersion` is used for enforcing minimum `-beta` versions; `minimumToolbeltVersion` is used for enforcing minimum stable version. You'll probably want to maitain the property `minimumToolbeltPrereleaseVersion <= minimumToolbeltVersion`.
```
lariciamota, line 30:
```suggestion
The `minimumToolbeltPrereleaseVersion` is used for enforcing minimum `-beta` versions; `minimumToolbeltVersion` is used for enforcing minimum stable version. You'll probably want to maintain the property `minimumToolbeltPrereleaseVersion <= minimumToolbeltVersion`.
```

docs/MAINTAINING.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Deprecating a version
+
+Toolbelt has a way of enforcing users to not use deprecated versions. From time to time toolbelt checks the `npm` registry checking if the current version is deprecated - if it's the user won't be allowed to run commands and will be presented a message like the following:
+
+```
+This version 2.93.0 was deprecated. Please update to the latest version: 'yarn global add vtex'.
+```
+
+This enforcement is on the client side and can be disabled on a execution by setting an environment variable (check check [EnvVariables](../src/lib/constants/EnvVariables.ts)).
```
lariciamota, line 60:
```suggestion
This enforcement is on the client side and can be disabled on an execution by setting an environment variable (check [EnvVariables](../src/lib/constants/EnvVariables.ts)).
```

docs/MAINTAINING.md
```diff
@@ -0,0 +1,76 @@
...
+
+That's it, now you can check on splunk the latest config changes (all changes are logged):
+
+```
+index=io_vtex_logs app=vtex.toolbelt-config-server@* data.event!=NULL workspace=master
+```
+
+### Deprecating a version
+
+Toolbelt has a way of enforcing users to not use deprecated versions. From time to time toolbelt checks the `npm` registry checking if the current version is deprecated - if it's the user won't be allowed to run commands and will be presented a message like the following:
```
lariciamota, line 54:
```suggestion
Toolbelt has a way of enforcing users to not use deprecated versions. From time to time toolbelt access the `npm` registry to check if the current version is deprecated - if it's the user won't be allowed to run commands and will be presented a message like the following:
```

docs/MAINTAINING.md
```diff
@@ -0,0 +1,76 @@
...
+```
+npm deprecate vtex@2.92.0 -m "Message explaining why this version was deprecated, for example: Bug when sending files to link"
+```
+
+2. Change the `latest` tag to the previous version (you'll have to check it):
+```
+npm dist-tag add vtex@2.91.0 latest
+```
+
+3. Make sure the we aren't enforcing the deprecated version (see [forcing a new version](##forcing-a-new-version))
```
lariciamota, line 74: Maybe that's what you wanted?
```suggestion
3. Make sure that we aren't enforcing the deprecated version (see [forcing a new version](##forcing-a-new-version))
```

docs/MAINTAINING.md
VictoriaMolgado: Napoli sends me more context on Slack, and I maybe we can introduce a "Why minimumToolbeltVersion". Just three lines, won't be a huge message for readers 👀  

Suggested copy: 

"Developers have several benefits when using the updated version of our CLI. That's why we push the minimumToolbeltVersion to then: s we can ensure they won't face security bugs that may be already resolved, get our developer experiences improves, and even help us more sending use metrics."

docs/MAINTAINING.md
```diff
@@ -0,0 +1,84 @@
+## Maintaining VTEX Toolbelt
+
+This document is for people working on VTEX as Toolbelt maintainers.
+
+### Forcing a new version
+
+Toolbelt has a way of enforcing users to use a minimum version choosed by us. This enforcement is done on the client side (can be disabled by an environment variable, check [EnvVariables](../src/lib/constants/EnvVariables.ts)) and on `vtex.builder-hub` - users using an older toolbelt version will not be allowed to link or publish. This feature is useful for enforcing users to update to the latest version, which can benefit users in many situations, like the following:
+
+- Bugs on toolbelt usage were fixed, providing a better experience.
```
VictoriaMolgado, line 9: That's a good way to communicate why the minimumToolbeltVersion is important 

LGTM! :D

PR #930, opened 2020-08-10
Fix new login flow

tiagonapoli:
There were two issues with the new login flow, this PR solves them:
- For some reason the previous package used to get a free port wasn't working properly on some OSs - The package was then changed from `get-port` to `detect-port`.
- In WSL2 the address `127.0.0.1:port` wasn't reachable. This was solved by specifying the hostname on `server.listen`.

Even though changing the address solved the problem for WSL2, some older WSL2 versions will still have this issue, and the solution is update to Windows to the latest version (at least 19041.421). However a reliable way of detecting if the user is on WSL2 or WSL1 is not that easy and because of this as a quick fix I decided to warn all WSL users about possible login issues regarding outdated windows versions.

```
yarn global add vtex@2.107.1-beta.1
```
Check if `vtex login` works properly.

#### Types of changes

#### Chores checklist
- [X] Update `CHANGELOG.md`


src/api/error/ErrorKinds.ts
```diff
@@ -16,4 +16,5 @@ export const ErrorKinds = {
   APP_LOGS_PARSE_ERROR: 'LogsParseError',
   STICKY_HOST_ERROR: 'StickyHostError',
   FLOW_ISSUE_ERROR: 'FlowIssue',
+  LOGIN_SERVER_START_ERROR: 'LoginServerStartError',
```
tiagonapoli, line 19: I added an specific error for login server startup

src/lib/auth/AuthProviders/OAuthAuthenticator/LoginServer.ts
```diff
@@ -0,0 +1,185 @@
...
+  public setLoginState(val: string) {
+    this.loginState = val
+  }
+
+  public start() {
+    return asyncRetry(
+      async (bail, attemptNumber) => {
+        try {
+          // detectPort will get the specified port or, if it's in use, another ramdom unnused port
+          this.port = await detectPort(3000)
```
tiagonapoli, line 76: Now `detectPort` is used instead of `get-port`

src/lib/auth/AuthProviders/OAuthAuthenticator/LoginServer.ts
```diff
@@ -0,0 +1,185 @@
...
+      const server = this.app.listen(port, LoginServer.HOSTNAME, () => {
+        server.on('connection', socket => {
+          socket.unref()
+        })
+
+        server.removeListener('error', reject)
+        resolve(server)
+      })
+
+      server.on('error', reject)
```
tiagonapoli, line 116: There was an error in the previous version - I was using `this.app.on('error', reject)`, and errors on server startup will be emitted by the server, not the app

src/lib/auth/AuthProviders/OAuthAuthenticator/LoginServer.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  public close() {
+    this.server.unref()
+    this.server.close()
+  }
+
+  private initServer(port: number): Promise<Server> {
+    return new Promise((resolve, reject) => {
+      const server = this.app.listen(port, LoginServer.HOSTNAME, () => {
```
tiagonapoli, line 107: Now instead of `this.app.listen(port, callback)` we're specifying the host

src/lib/auth/AuthProviders/OAuthAuthenticator/index.ts
```diff
@@ -0,0 +1,122 @@
...
+          `If you do, make sure your Windows is up to date and try again (${ColorifyConstants.URL_INTERACTIVE(
+            'https://support.microsoft.com/en-us/help/4027667/windows-10-update'
+          )}).`
+        )
+
+        logger.warn(
+          `In case login errors persist after updating please create an issue on ${ColorifyConstants.URL_INTERACTIVE(
+            'https://github.com/vtex/toolbelt/issues'
+          )}. We'll promptly help you finding a solution.`
+        )
```
tiagonapoli, line 60: These warnings were added

src/lib/auth/AuthProviders/OAuthAuthenticator/index.ts
```diff
@@ -0,0 +1,122 @@
...
+      const url = await this.loginUrl(account, loginState)
+      opn(url, { wait: false })
+
+      if (isWsl) {
+        logger.warn(
+          "We noticed you're using WSL, in which case you may face login issues depending on your WSL version."
+        )
+
+        logger.warn(
+          `If you do, make sure your Windows is up to date and try again (${ColorifyConstants.URL_INTERACTIVE(
```
lariciamota, line 51: I don't know if it was intentional to show the whole URL, but if not you could use this [function](https://github.com/vtex/toolbelt/blob/59ce5fe364b6734f6619301f4194abb0e08e8acd/src/lib/utils/Messages.ts#L5) to show hyperlink

src/lib/auth/AuthProviders/OAuthAuthenticator/index.ts
```diff
@@ -0,0 +1,122 @@
+import axios from 'axios'
+import { createHash } from 'crypto'
+import isWsl from 'is-wsl'
+import jwt from 'jsonwebtoken'
+import opn from 'opn'
```
lariciamota, line 5: This package has been deprecated and renamed to [open](https://www.npmjs.com/package/open).
I think it would be better if we updated it! :)

src/lib/auth/AuthProviders/OAuthAuthenticator/index.ts
```diff
@@ -0,0 +1,122 @@
...
+        logger.warn(
+          `If you do, make sure your Windows is up to date and try again (${ColorifyConstants.URL_INTERACTIVE(
+            'https://support.microsoft.com/en-us/help/4027667/windows-10-update'
+          )}).`
+        )
+
+        logger.warn(
+          `In case login errors persist after updating please create an issue on ${ColorifyConstants.URL_INTERACTIVE(
+            'https://github.com/vtex/toolbelt/issues'
+          )}. We'll promptly help you finding a solution.`
```
lariciamota, line 59: This way sounds better:
```suggestion
          )}. We'll promptly help you find a solution.`
```
but I'm not 100% sure 🤔, what do you think?

src/lib/auth/AuthProviders/OAuthAuthenticator/index.ts
```diff
@@ -0,0 +1,122 @@
...
+      opn(url, { wait: false })
+
+      if (isWsl) {
+        logger.warn(
+          "We noticed you're using WSL, in which case you may face login issues depending on your WSL version."
+        )
+
+        logger.warn(
+          `If you do, make sure your Windows is up to date and try again (${ColorifyConstants.URL_INTERACTIVE(
+            'https://support.microsoft.com/en-us/help/4027667/windows-10-update'
```
lariciamota, line 52: Is it the minimum allowed?

src/lib/auth/AuthProviders/OAuthAuthenticator/index.ts
tiagonapoli: Actually I tried some months ago to use the new package, but we had some issues with some systems - WSL - I made a contribution to fix the issue https://github.com/sindresorhus/open/pull/171 - however we began having issues with other systems and some specific accounts URLs :(
Because of this I decided to wait for the 'open' package to evolve more and then update

src/lib/auth/AuthProviders/OAuthAuthenticator/index.ts
tiagonapoli: I couldn't think of a text to use as the URL :( - and there's an issue with WSL -> as of now it doesnt support ansi url codes (I tested on WSL1 and WSL2, Ubuntu 20.04 and 18.04)

However if you have a suggestion to use as a text I would very much accept it :)

src/lib/auth/AuthProviders/OAuthAuthenticator/index.ts
tiagonapoli: Oops, didn't understand the question

src/lib/auth/AuthProviders/OAuthAuthenticator/index.ts
tiagonapoli: I think it sounds better indeed

src/lib/auth/AuthProviders/OAuthAuthenticator/index.ts
lariciamota: This function previously checks the hyperlink support, would it be enough to solve this problem?

Well, I can only think in something simple like `here`, maybe @VictoriaMolgado can think about a better option

src/lib/auth/AuthProviders/OAuthAuthenticator/index.ts
lariciamota: Sorry, nevermind, I'd misunderstood it

src/lib/auth/AuthProviders/OAuthAuthenticator/index.ts
tiagonapoli: The only problem here is that we know in this case the function output would be like the following, because of WSL:
```
If you do, make sure your Windows is up to date and try again. For update instructions check here (URL).
```
In this case, what you would prefer?
```
If you do, make sure your Windows is up to date and try again. For update instructions check here (URL).
```
or
```
If you do, make sure your Windows is up to date and try again. For update instructions check URL.
```

lariciamota: We are inside the `if wsl`, for a moment I forgot about it 😆  Indeed, it looks better this way

PR #933, opened 2020-08-11
Refact/workspace plugin

VerasThiago:
Transform `vtex workspace` into a plugin command

#### Types of changes
- [x] Refactor (non-breaking change that only makes the code better)

#### Chores checklist
- [x] Update `CHANGELOG.md`


src/api/modules/workspace/create.ts
```diff
@@ -103,6 +103,12 @@ export const workspaceCreator: WorkspaceCreator = async ({
 
     return 'created'
   } catch (err) {
+    if (err.response?.data.code === 'ArgumentNull') {
+      throw createFlowIssueError(
```
VerasThiago, line 107: @VictoriaMolgado helped me about this new error flow to the user, what do you guys think about it ?

CHANGELOG.md
```diff
@@ -5,13 +5,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+- [vtex workspace] Refactor to use as oclif plugin.
 
 ## [2.108.1] - 2020-08-12
 
 ### Changed
 - [vtex login] Use new VTEX ID login flow:
-  - Fix free TCP port retrieval: use `detect-port` package instead of `get-port`.
-  - Fix issue on WSL2 by specifying host binding to `127.0.0.1` when starting the server.
+- Fix free TCP port retrieval: use `detect-port` package instead of `get-port`.
+- Fix issue on WSL2 by specifying host binding to `127.0.0.1` when starting the server.
```
tiagonapoli, line 16: In this case you can maintain the additional spaces (they are subtopics)

src/api/lib/constants/Messages.ts
```diff
@@ -0,0 +1,44 @@
+import chalk from 'chalk'
+import { formatHyperlink } from '../../../lib/utils/Messages'
+import { ColorifyConstants } from './Colors'
+
+export const Messages = {
```
tiagonapoli, line 5: I think messages are command specific - maybe each command should have its own

src/api/index.ts
```diff
@@ -11,5 +11,6 @@ export * from './locator'
 export * from './modules'
 export * from './table'
 export * from './verbose'
-export { BatchStream } from './typings'
+export * from './lib'
```
tiagonapoli, line 14: How about we move the ./lib/constants to ./constants? Then here we would export `./constants'

src/api/modules/workspace/create.ts
```diff
@@ -103,6 +103,12 @@ export const workspaceCreator: WorkspaceCreator = async ({
...
+        `You need to pick a name for the new workspace. Run vtex worspace create [WORKSPACENAME]. You can name it using your name with a number, for example vtex workspace create johndoe01`
```
tiagonapoli, line 108:
```suggestion
        `You need to pick a name for the new workspace. Run vtex workspace create [WORKSPACENAME]. You can name it using your name with a number, for example vtex workspace create johndoe01`
```

src/api/modules/workspace/create.ts
tiagonapoli: Should we add some colors highlighting the code parts, or '' enclosing them? In this part for example:
```
Run vtex workspace create [WORKSPACENAME]
```
Should we print?
```
Run 'vtex workspace create [WORKSPACENAME]'
```
Should we add colors in it?

VerasThiago: Good ideia

tiagonapoli: Nicee, what do you think abouot the last one?
`for example vtex workspace create johndoe01`
Should we color it too?

tiagonapoli: Or maybe just enclose it in ''

src/api/constants/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './Messages'
```
tiagonapoli, line 1: I think messages could be moved back out of api, what do you think?
Since each command will have its own messages maybe we shouldn't share this

src/api/modules/workspace/create.ts
VerasThiago: Option 1,2 or 3 ?

lariciamota: I think those should be VTEX pink, since now we have defined colors for our CLI, you can see it in [here](https://www.figma.com/file/GpNy0htTpJ7uvw7WKqGdLH/Toolbelt-DX-review?node-id=373%3A89) and they are on `ColorifyConstants` 😄

VictoriaMolgado: Use the green color for now :) 
That's because we want to use the vtex pink only for commands. Checkout this proto-guide: https://www.notion.so/Toolbelt-Styleguide-1943920c5cb84d4a9e22a501c43beab2#0f28a0efe490471f9aca3424d3a01936

Is it possible to associate a class to the line to change it easily if it makes sense in the future?

src/api/modules/workspace/create.ts
```diff
@@ -103,6 +103,16 @@ export const workspaceCreator: WorkspaceCreator = async ({
 
     return 'created'
   } catch (err) {
+    if (err.response?.data.code === 'ArgumentNull') {
+      throw createFlowIssueError(
+        `You need to pick a name for the new workspace. Run ${chalk.green(
+          'vtex workspace create [WORKSPACENAME]'
+        )}. You can name it using your name with a number, for example ${chalk.green(
+          `vtex workspace create johndoe01`
+        )}`
```
lariciamota, line 112: What about adding it to the messages file?

lariciamota: And using the colors from ColorifyConstants

VerasThiago: Good idea

PR #940, opened 2020-08-27
Create commands docs

tiagonapoli:
There are two alternatives:
- Topic separated files
- All commands in a single file

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -1,0 +1,1294 @@
+# Commands
+
+* [`vtex add APPID [ITHAPPID]`](#vtex-add-appid-ithappid)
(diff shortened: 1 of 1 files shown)


VictoriaMolgado:
I believe that the "All commands in a single file" approach is a path more coherent with our guidelines at this moment (centralize information and being more informative by now, until the moment we discover which information to prioritize, etc). 

Once the all-commands possibility comes with a "summary", that makes ease to find what the user is looking for, let's use this approach by now as a quick win (as you sad at Slack) and continue discovering what to do next but with a more organized base from now 😅 

We can add a list with all commands to the README file.

PR #950, opened 2020-09-10
Feat/new installer

VerasThiago:
#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- [installer] New deploy method by using `oclif-dev pack` and `oclif-dev publish`
```
tiagonapoli, line 9: Not sure about the `by` usage here: 🤔 
```suggestion
- [installer] New deploy method using `oclif-dev pack` and `oclif-dev publish`.
```

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+
+### Changed
+- [tsconfig] Changed build folder name from `build` to `lib`
```
tiagonapoli, line 12: nit
```suggestion
- [tsconfig] Changed build folder name from `build` to `lib`.
```

package.json
```diff
@@ -51,6 +51,7 @@
     "@oclif/command": "^1.0.0",
     "@oclif/config": "^1.0.0",
     "@oclif/plugin-help": "^2.0.0",
+    "@thiagoveras/plugin-update": "^1.0.0",
```
tiagonapoli, line 54: In the future we can fork this project under vtex name 🚀 
(I'm in the wrong here, because of `@tiagonapoli/oclif-plugin-spaced-commands` - it was supposed to be a temporary solution because of toolbelt 3.0 😢 )

src/modules/release/utils.ts
```diff
@@ -44,10 +44,10 @@ export class ReleaseUtils {
     return version
   }
 
-  public incrementVersion = (rawOldVersion: string, releaseType: string, tagName: string) => {
+  public incrementVersion = (rawOldVersion: string, releaseType: semver.ReleaseType, tagName: string) => {
     const oldVersion = semver.valid(rawOldVersion, true)
     if (tagName !== 'stable' && releaseType !== 'prerelease') {
-      return semver.inc(oldVersion, `pre${releaseType}`, tagName)
+      return semver.inc(oldVersion, `pre${releaseType}` as semver.ReleaseType, false ,tagName)
```
tiagonapoli, line 50: Were you able to test this? I think a `vtex release` on an app would be sufficient

scripts/uninstall.sh
```diff
@@ -0,0 +1,38 @@
...
+
+
+    # run inside sudo
+    $SUDO bash <<SCRIPT
+    set -e
+
+    echoerr() { echo "\$@" 1>&2; }
+
+    if [[ ! ":\$PATH:" == *":/usr/local/bin:"* ]]; then
+        echoerr "You don't have /usr/local/bin path. So your CLI are not installed"
```
tiagonapoli, line 21: Maybe:
```suggestion
        echoerr "You don't have /usr/local/bin path. VTEX CLI is not installed"
```

scripts/uninstall.sh
```diff
@@ -0,0 +1,38 @@
...
+        exit 1
+    fi
+
+    echo "Uninstalling VTEX"
```
tiagonapoli, line 25: Maybe:
```suggestion
    echo "Uninstalling VTEX CLI"
```

scripts/uninstall.sh
```diff
@@ -0,0 +1,38 @@
...
+    rm -rf /usr/local/bin/vtex
+    echo "removing command -v vtex ..."
+    rm -f \$(command -v vtex) || true
+    echo "removing ~/.local/share/vtex ..."
+    rm -rf ~/.local/share/vtex
+    echo "removing /usr/local/lib/vtex ..."
+    rm -rf /usr/local/lib/vtex
+    
+SCRIPT
+    echo "VTEX Uninstalled"
```
tiagonapoli, line 37: Maybe: (we can confirm with molgado if we should call toolbelt "VTEX CLI")
```suggestion
    echo "VTEX CLI Uninstalled"
```

package.json
VerasThiago: This is just a testing plugin, the final result will be under vtex name

src/CLIPreTasks/OutdatedChecker/OutdatedChecker.ts
```diff
@@ -23,11 +24,18 @@ export class OutdatedChecker {
...
+      `${chalk.bold(
+        `Your Toolbelt version (${pkgJson.version}) is outdated`
+      )}. You must update using the same method you used to install. Here are some examples:` +
+      `\n\n` +
+      `• If you installed using ${ColorifyConstants.COMMAND_OR_VTEX_REF(
+        `yarn`
+      )}, update running ${ColorifyConstants.COMMAND_OR_VTEX_REF(`yarn global add vtex`)}.` +
+      `\n\n` +
+      `• If you installed using our new method there is in alpha-version, update running ${ColorifyConstants.COMMAND_OR_VTEX_REF(
+        `vtex autoupdate`
```
dhasuda, line 37: This `errMsg` is duplicates. Can you define it in only one place, so it's easier to change it if needed?

dhasuda (changes requested):
The error message for installing using the same method is defined in many places, can you define the message in only one place?

src/CLIPreTasks/OutdatedChecker/OutdatedChecker.ts
VerasThiago: Actually is not the same, only differ by `deprecated` or `outdated`

VerasThiago: Maybe creating a base message and receive an argument of the status is a better solution, what do you think about it?

dhasuda: Think its good! If there is a new way of installing it, for instance, it will make the refactor of this text much easier

dhasuda (approved):
LGTM

PR #953, opened 2020-09-18
Feature/billing options terms

marlonguimaraes:
### Some context about what already exists

VTEX accounts can install apps using our CLI/toolbelt by using the command `vtex install`. Some of those apps have `billingOptions` (an optional field in the `manifest.json` file that describes how/if the app is paid). Apps without `billingOptions` are not relevant to this PR.

An app with `billingOptions` can be free or paid. In both cases, when the user tries to install (by using `$ vtex install`) an app, she is informed by the toolbelt that the app is free or paid. Also, all of these apps can optionally have a "Terms of Service" URL, which the user can visit to see the app's terms.

**The case of paid apps**

If the app is paid, it's shown to the user how exactly she is going to be possibly charged by the app. There are only two ways the user can be charged for using an app:

1) A monthly subscription price in reais (BRL, R$), like Netflix and Spotify.
2) A variable ammount according to some metric value. For example, the app `sms-provider` charges `R$ 0.07` for each SMS sent by the account that installed the app. If the account sent no SMSs, the account will not be charged at all. If it sent 10 SMS, it will be charged `R$ 0.70` at the end of the month.

Note that can 1) or 2) can both exist simultanously. 

### Who is the user or who are we talking to?

The user is anyone logged in a VTEX account and is interested in installing free or paid apps via the CLI/toolbelt.

### What does this PR changes?

1) Deal with the new `billingOptions` format (doens't impact the UI)
2) Words and colors that is shown to that is trying to install an app. No new information was added, we're just trying to make it cleaner  (impacts UI)

You can check how the UI is now by using the commands `vtex install vtex.sms-provider` or `vtex install yourviews.yourviewsreviews` (omitted here because there are a lot of pictures already, but hit me up on Slack if u need pictures).

1)  Free app without license file

```sh
vtex-test install yourviews.yourviewsreviews
```

2) Free app with license file (the URL https://extensions.myvtex.com/_v/terms/vtex.google-customer-reviews@x should work)
```sh
vtex-test install vtex.google-customer-reviews
```

3) Paid app with new `billingOptions` and license files:

```sh
vtex-test install vtex.google-shopping
```

4) One metric 

```sh
vtex-test install vtex.sms-provider
```

5) Should show subscription price and multiple metrics:

```
vtex-test install pilateslovers.billingoptionstest@0.x
```

1) 
image

2)
image

3)
image

4)
image

5)
image

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`


src/api/modules/utils.ts
```diff
@@ -129,81 +128,99 @@ export const appIdFromRegistry = (app: string, majorLocator: string) => {
...
+  })
+}
+
+const billingInfoFromPolicy = ({ currency, billing: { items } }: Policy): BillingInfo => {
+  const subscription = items.reduce<number>((sum, { fixed }) => {
+    if (fixed) {
+      sum += fixed
+    }
+    return sum
+  }, 0)
```
pedromtec, line 182:
```suggestion
  const subscription = items.reduce<number>(
  (sum, { fixed }) => (fixed ? sum + fixed : sum),
  0
)
```

pedromtec (changes requested):

src/api/modules/utils.ts
```diff
@@ -129,81 +128,99 @@ export const appIdFromRegistry = (app: string, majorLocator: string) => {
     .catch(handleError(app))
 }
 
-export function optionsFormatter(billingOptions: BillingOptions) {
+const FREE_BILLING_OPTIONS_TYPE = 'free'
+export const isFreeApp = (type?: string, free?: boolean) => type === FREE_BILLING_OPTIONS_TYPE || free
```
arturpimentel, line 132: You could also receive the billing options directly
```suggestion
export const isFreeApp = ({ type, free }: BillingOptions) => type === FREE_BILLING_OPTIONS_TYPE || free
```

src/modules/apps/install.ts
```diff
@@ -24,11 +25,32 @@ const promptPolicies = async () => {
...
+      return false
+    }
+    throw err
+  }
+}
+
+const licenseURL = async (app: string, termsURL?: string): Promise<string | undefined> => {
+  const [name, argVersion] = app.split('@')
+  const version = argVersion ?? 'x'
+  return (await hasLicenseFile(name, version)) ? `https://extensions.myvtex.com/_v/terms/${name}@${version}` : termsURL
```
arturpimentel, line 44:
```suggestion
return (await hasLicenseFile(name, version)) ? `https://apps.vtex.com/_v/terms/${name}@${version}` : termsURL
```

arturpimentel (commented):
Made just a few comments

src/api/modules/utils.ts
marlonguimaraes: I did that way befefore, although the code looks nicer I think it makes maintainability worse IMHO. If you want to search for all uses of "billingOptions" you would find that line, but that function doesn't use the whole billingOptions

src/modules/apps/install.ts
marlonguimaraes: Done in 6dcee75b3c884f4f750ee7ca20680fff37b73dff, thanks!

src/api/modules/utils.ts
marlonguimaraes: Nice, done in 6dcee75b3c884f4f750ee7ca20680fff37b73dff, thanks!

src/api/modules/utils.ts
```diff
@@ -129,81 +128,94 @@ export const appIdFromRegistry = (app: string, majorLocator: string) => {
...
   const table = new Table({
     head: [{ content: chalk.cyan.bold('Billing Options'), colSpan: 2, hAlign: 'center' }],
     chars: { 'top-mid': '─', 'bottom-mid': '─', 'mid-mid': '─', middle: ' ' },
   })
-
-  if (billingOptions.free) {
-    table.push([{ content: chalk.green('This app is free'), colSpan: 2, hAlign: 'center' }])
+  const billingInfo = buildBillingInfo(plans, policies)
+  if (isFreeApp(type, free) || !billingInfo) {
+    table.push([{ content: chalk.green('Free app'), colSpan: 2, hAlign: 'center' }])
```
VerasThiago, line 212: How about using the colors from [Colors.ts](https://github.com/vtex/toolbelt/blob/60bb7edd56/src/api/constants/Colors.ts) ? 
Ex: ```chalk.hex(COLORS.GREEN)('Free app')```
And probably add new `ColorifyConstants` `readonly`  ? (I'm not sure about this, just a possibility)

src/api/modules/utils.ts
```diff
@@ -129,81 +128,94 @@ export const appIdFromRegistry = (app: string, majorLocator: string) => {
...
-          content: `The monthly amount will be charged in ${chalk.red(policy.currency)}`,
-          colSpan: 2,
-          hAlign: 'center',
-        },
-      ])
-    })
+    chalkBillingTable(table, billingInfo)
+  }
+  if (licenseURL && licenseURL.length > 0) {
+    table.push([{ content: chalk.yellow.bold('App Terms of Service: ') }, { content: licenseURL }])
```
VerasThiago, line 217: Same here

src/modules/apps/install.ts
```diff
@@ -24,11 +25,32 @@ const promptPolicies = async () => {
...
+  const [name, argVersion] = app.split('@')
+  const version = argVersion ?? 'x'
+  return (await hasLicenseFile(name, version)) ? `https://apps.vtex.com/_v/terms/${name}@${version}` : termsURL
+}
+
 const checkBillingOptions = async (app: string, billingOptions: BillingOptions, force: boolean) => {
+  const { termsURL, type, free } = billingOptions
+  const license = await licenseURL(app, termsURL)
   log.warn(
     `${chalk.blue(app)} is a ${
```
VerasThiago, line 51: Same here about the`Colors.ts`

src/api/modules/utils.ts
```diff
@@ -129,81 +128,94 @@ export const appIdFromRegistry = (app: string, majorLocator: string) => {
     .catch(handleError(app))
 }
 
-export function optionsFormatter(billingOptions: BillingOptions) {
+const FREE_BILLING_OPTIONS_TYPE = 'free'
+export const isFreeApp = (type?: string, free?: boolean) => type === FREE_BILLING_OPTIONS_TYPE || free
+
+type BillingInfo = {
```
VerasThiago, line 134: If this type will be used by more files like `BillingOptions`, maybe is a good idea to move to `/src/typings/global.d.ts`

src/modules/apps/install.ts
```diff
@@ -24,11 +25,32 @@ const promptPolicies = async () => {
...
-      billingOptions.free ? chalk.green('free') : chalk.red('paid')
-    } app. To install it, you need to accept the following Terms:\n\n${optionsFormatter(billingOptions)}\n`
+      isFreeApp(type, free) ? chalk.green('free') : chalk.red('paid')
```
VerasThiago, line 52: Other idea is to have a `Messages.ts` file like [Workspace Command](https://github.com/vtex/cli-plugin-workspace/blob/master/src/lib/constants/Messages.ts), in order to separate messages from code and group all of them.

src/api/modules/utils.ts
arturpimentel: I see. I just thought It could be used as something like "give me your billing options and I'll find out if this is a free app". One middle ground I know is do this:
```suggestion
export const isFreeApp = ({ type, free }: { type?: string, free?: boolean }) => type === FREE_BILLING_OPTIONS_TYPE || free
```
Then, you can call this by providing the billing options, while also being specific with what you really care about

marlonguimaraes: Nice, thats better! Done in 25283acc92ee221ef8640460eb39d837c48b5692 , thanks!

src/modules/apps/install.ts
marlonguimaraes: Very elegant idea, thanks for sharing! Done in 1a1f01e7e948ad1dcbe61c6a7cc761f9a28c6dd4

src/api/modules/utils.ts
marlonguimaraes: Currently only used here

src/modules/apps/install.ts
marlonguimaraes: Done in 1a1f01e7e948ad1dcbe61c6a7cc761f9a28c6dd4, thanks!

src/api/modules/utils.ts
marlonguimaraes: Done in 1a1f01e7e948ad1dcbe61c6a7cc761f9a28c6dd4, thanks!

src/api/modules/utils.ts
marlonguimaraes: Done in 1a1f01e7e948ad1dcbe61c6a7cc761f9a28c6dd4, thanks!

arturpimentel (approved):

src/api/modules/billingMessages.ts
```diff
@@ -0,0 +1,21 @@
+import chalk from 'chalk'
```
VerasThiago, line 1: Since this file is not being used by other files, I think that he don't belong to `/api` folder. How about moving to `/src/lib/constant/`

marlonguimaraes: Makes sense, done in c3a67717ae2ec0c3ee9d60efbd6a701ab487cc86 !

pedromtec (approved):

VictoriaMolgado (changes requested):
Nice work!
Thanks for the context. It's gonna be great to share with the `dev tools team` the news on the CLI. 

Here are some design inputs:

- The "Billing options" title must be white. It is not interactive and it stands out enough for being at the top of a table. 
- We should always use the "column labels". Te first should be "App", the same term we are using to talk about what the user is installing
- The second column, "Pricing"
- When it's free, we say "Free app" under "Pricing"
- The "Free app" text color os white (it's not interactive and the context and table will provide the visual hierarchy to the user to understand the information)

**Why use this format?** 
We apply a pattern that the user "learns to read": she will always use the columns-titles as an anchor that makes ease to know "what is that information", with time it becomes automatic reading. 

**About the messages**
To the paid apps, I don't really get what is happening at the message. It just does not display the "its a free app"? We should have a specific message for each case. As the following:

Free app: 
- Remove the "warn" tag;
- Use the message "`app-name` is a free app! Accept the app's terms to install it."

Paid: 
- Remove the "warn" tag: 
- Message: "Here are some details about the billing terms of `app-name`. To install it, you need to accept the app's terms."

The overall visual presentation looks good to me. The messages fixies will impact the overall quality of Toolbelt's communication. 

Many thanks for the design attention
If you want, you can read about our guidelines here > https://github.com/vtex/ux-writing/blob/new-content/docs/text-patterns/toolbelt-cli.md

pedromtec (approved):

VictoriaMolgado (approved):
LGTM!
Thanks for the dedication 🤗

VerasThiago (approved):
LGTM

PR #966, opened 2020-10-09
Change version update notifier structure

VerasThiago:
`updateNotify` method was not updating correctly `toolbelt` versions based on `channels`

Now `toolbelt` will notify the CLI version update based on his own `channel`

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/src/modules/utils.ts
+++ b/src/modules/utils.ts
@@ -60,6 +60,22 @@ const cleanVersion = (appId: string) => {
   )(appId)
 }
 
+// Return version tag
+// Example: 2.115.0-beta.somehash   -> beta
+// Example: 2.115.0                 -> latest
+export const getDistTag = (version: string) => {
+  const distTag = version.split('-')[1]
+  return distTag ? distTag.split('.')[0] : 'latest'
+}
+
+// Return version and tag only
+// Example: 2.115.0-beta.somehash   -> 2.115.0-beta
+// Example: 2.115.0                 -> 2.115.0
+export const getSimpleVersion = (_version: string) => {
+  const [version, distTag] = _version.split('-')
+  return version.concat(distTag ? '-'.concat(distTag.split('.')[0]) : '')
+}
+
 export const matchedDepsDiffTable = (title1: string, title2: string, deps1: string[], deps2: string[]) => {
   const depsDiff = diffArrays(deps1, deps2)
   // Get deduplicated names (no version) of the changed deps.
--- a/src/modules/utils.ts
+++ b/src/modules/utils.ts
@@ -64,16 +64,17 @@ const cleanVersion = (appId: string) => {
 // Example: 2.115.0-beta.somehash   -> beta
 // Example: 2.115.0                 -> latest
 export const getDistTag = (version: string) => {
-  const distTag = version.split('-')[1]
-  return distTag ? distTag.split('.')[0] : 'latest'
+  const regex = /(?:-([0-9A-Za-z-]*))/g
+  const distTag = version.match(regex)
+  return distTag ? distTag[0].substring(1) : 'latest'
 }
 
 // Return version and tag only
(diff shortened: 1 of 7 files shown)


src/modules/utils.ts
```diff
@@ -60,6 +60,22 @@ const cleanVersion = (appId: string) => {
...
+  return distTag ? distTag.split('.')[0] : 'latest'
+}
+
+// Return version and tag only
+// Example: 2.115.0-beta.somehash   -> 2.115.0-beta
+// Example: 2.115.0                 -> 2.115.0
+export const getSimpleVersion = (_version: string) => {
+  const [version, distTag] = _version.split('-')
+  return version.concat(distTag ? '-'.concat(distTag.split('.')[0]) : '')
+}
```
dhasuda, line 77: _suggestion_ (non blockin)

You can also use Regex to perform this operation
`/([0-9]+\.[0-9]+\.[0-9]+)/g`

You can use [the `match` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match)

src/update.ts
```diff
@@ -4,34 +4,24 @@ import { ColorifyConstants } from './api/constants/Colors'
...
 
 import * as pkg from '../package.json'
+import { getDistTag, getSimpleVersion } from './modules/utils'
 
 export function updateNotify() {
-  const notifier = updateNotifier({ pkg, updateCheckInterval: 1000 * 60 * 60 * 1 })
-  if (notifier.update && notifier.update.latest !== pkg.version) {
-    const oldVersion = notifier.update.current
+  const distTag = getDistTag(pkg.version)
+  const notifier = updateNotifier({ pkg, distTag, updateCheckInterval: 1000 * 60 * 60 * 1 })
```
dhasuda, line 11: Update `1000 * 60 * 60 * 1` with a more verbose variable name. I'm not sure what's the unit of measure, but let's say this means one hour:
```
const ONE_HOUR = 1000 * 60 * 60 * 1
...
const notifier = updateNotifier({ pkg, distTag, updateCheckInterval: ONE_HOUR })
```

src/update.ts
```diff
@@ -4,34 +4,24 @@ import { ColorifyConstants } from './api/constants/Colors'
...
+  if (notifier.update && notifier.update.latest !== getSimpleVersion(pkg.version)) {
```
dhasuda, line 12: _suggestion_
You can write:
```
if (notifier.update?.latest !== getSimpleVersion(pkg.version)) {...}
```

dhasuda (changes requested):
I have two requests:
- Can you take the opportunity to add some documentation on this if it doesn't already exists?
- I like the tests you've added! But is there any manual test we can perform as well?

src/modules/utils.ts
VerasThiago: Excellent idea!

src/update.ts
```diff
@@ -4,34 +4,26 @@ import { ColorifyConstants } from './api/constants/Colors'
 import { Messages } from './lib/constants/Messages'
 
 import * as pkg from '../package.json'
+import { getDistTag, getSimpleVersion } from './modules/utils'
 
 export function updateNotify() {
-  const notifier = updateNotifier({ pkg, updateCheckInterval: 1000 * 60 * 60 * 1 })
-  if (notifier.update && notifier.update.latest !== pkg.version) {
-    const oldVersion = notifier.update.current
+  const ONE_HOUR = 1000 * 60 * 60 * 1
```
dhasuda, line 10: _non blocking_
You can declare this variable outside the `updateNotify` function

dhasuda (approved):
LGTM! I've left only one non-blocking comment

src/update.ts
VerasThiago: I thought about it, but any other functions will use it.

PR #981, opened 2020-11-09
Fix/login open browser

VerasThiago:
Change `open` lib

This library is deprecated and in some Linux distributions when I do the vtex login, it doesn't open the browser.

#### Types of changes

#### Chores checklist
- [X] Update `CHANGELOG.md`

### DIFF
--- a/src/nps.ts
+++ b/src/nps.ts
@@ -1,6 +1,8 @@
 import enquirer from 'enquirer'
 import moment from 'moment'
+import open from 'open'
 import opn from 'opn'
+import { ToolbeltConfig } from './api/clients/IOClients/apps/ToolbeltConfig'
 
 import { getNextFeedbackDate, saveNextFeedbackDate } from './api/conf'
 import { promptConfirm } from './api/modules/prompts'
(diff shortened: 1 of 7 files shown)


src/nps.ts
```diff
@@ -41,7 +43,15 @@ export async function checkAndOpenNPSLink() {
...
-      opn(NPSFormURL, { wait: false })
+      const configClient = ToolbeltConfig.createClient()
+      const { featureFlags } = await configClient.getGlobalConfig()
+
+      if (featureFlags.FEATURE_FLAG_NEW_OPEN_PACKAGE) {
+        open(NPSFormURL, { wait: false })
+      } else {
+        opn(NPSFormURL, { wait: false })
+      } 
+
```
dhasuda: _non blocking suggestion_:
The if with `featureFlags.FEATURE_FLAG_NEW_OPEN_PACKAGE` is used in a lot of places. You can make a selector somewhere else... in the future, if the logic gets more complex (perhaps you may need to use a _switch_ IDK)
```
const openPackageSelector(withFeatureFlag, withoutFeatureFlag) {
  // Logic to get feature flags...
 if (featureFlags.FEATURE_FLAG_NEW_OPEN_PACKAGE) {
    return withFeatureFlag
  } else {
    return withoutFeatureFlag
  }
}
```

dhasuda (approved):
LGTM! Only one non blocking coment

src/nps.ts
VerasThiago: In my point of view, the use o `Feature Flags` in complex cases need to encapsulate the login in other functions, like [here](https://github.com/vtex/cli-plugin-deploy/blob/7e936891abd5a2cf0e77bd82216bc441a67cbf9a/src/commands/deploy.ts#L24), but in this case I don't think that needed. What do you think?

dhasuda: I don't think it's needed. But the suggestion is about creating a selector of the function that will be used. Something that analyses the feature flags and the possibilities of execution, and chooses one. The ideia is to simply separate the logic of the feature flag selection from the rest of the code

dhasuda: Anyway, it's not a blocking comment, feel free to move on

PR #1011, opened 2021-02-01
Don't throw error when switching to same account

dhasuda:
[Jira card](https://vtex-dev.atlassian.net/browse/DT-360?atlOrigin=eyJpIjoiMWI3MmNiNzNlZmIzNDJmOGI3NDIwNTQ3NzYxMWUwZTYiLCJwIjoiaiJ9)

Don't throw error when switching to same account.

If I use the command:

vtex switch `storecomponents/breno`

And I'm at `storecomponents/foobar`, toolbelt will throw an error saying that I'm already at storecomponents and won't switch me to the workspace breno.

The ideal would be to just to move me to the new workspace.

After the command `yarn watch`, you may try:
```bash
vtex-test switch vtex/master

vtex-test switch vtex/master
# This should not throw an error

vtex-test switch vtex/davi
# This should not throw an error either

vtex-test switch storecomponents/master
# This behavior won't change
```

Screen Shot 2021-02-01 at 5 44 06 PM

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## [3.1.2-beta] - 2021-02-04
 ### Changed
 - Don't throw error when switching to same account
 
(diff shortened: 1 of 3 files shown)


VerasThiago (approved):
LGTM

VerasThiago:
Just 2 non-blocking observations: 
1 - How about changing the `log` variable to `logger` ? Jus to follow the pattern that `toolbelt-plugins` are using (since have a lot of places that still use `log`, I was changing gradually)

2- How about changing the colors, i.e: `chalk.blue` to `toolbelt constant colors` ? Is located at `/src/api/constant`, with this we can have a default tone to all color and change all of them much easily.

PR #1029, opened 2021-02-19
Release v2.121.6

VerasThiago:


tibuurcio (approved):
I'm having problems developing locally because we have a release version v2.121.6 and the code on master branch is on v2.121.5

This triggers a deprecation check when running locally
image

I believe merging this PR will resolve this :)

LGTM

PR #1040, opened 2021-03-09
Fix/remove topic when empty

VerasThiago:
- `toolbelt` folder, run : `yarn watch`
- Terminal, run: `vtex-test`

BEFORE:

image

AFTER

image

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/src/oclif/hooks/utils.ts
+++ b/src/oclif/hooks/utils.ts
@@ -3,6 +3,7 @@ import indent from 'indent-string'
 import { COLORS } from '../../api'
 import { renderList } from '../../../node_modules/@oclif/plugin-help/lib/list'
 import RootHelp from '../../../node_modules/@oclif/plugin-help/lib/root'
+import {OTHER_GROUP_ID} from './constants'
 
 export interface CommandI {
   name: string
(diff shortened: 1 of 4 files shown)


src/oclif/hooks/utils.ts
```diff
@@ -36,10 +36,13 @@ export function renderCommands(commandsId: Record<number, string>, groups: Comma
   body.push(help.root())
   body.push(' ')
 
-  for (const [key, value] of Object.entries(commandsId)) {
-    body.push(chalk.bold(value))
-    body.push(indent(renderCommand(groups[key !== '255' ? key : commandsGroupLength - 1], ctx), 2))
-    body.push(' ')
+  for (let [key, value] of Object.entries(commandsId)) {
+    key = key !== '255' ? key : (commandsGroupLength - 1).toString()
```
grisvtex, line 40: Please, extract `255` to a constant. Maybe `OTHERS_GROUP_ID`? What do you think?

src/oclif/hooks/utils.ts
```diff
@@ -36,10 +36,13 @@ export function renderCommands(commandsId: Record<number, string>, groups: Comma
...
+    if (groups[key].length) {
```
grisvtex, line 41: Please test explicity that the list length is greater than 0. Also This conditional is too optimistic. Maybe test first if `group[key]` is indeed different than `undefined`.

grisvtex (changes requested):

src/oclif/hooks/utils.ts
VerasThiago: Sure!

src/oclif/hooks/utils.ts
VerasThiago: You’re right!

VerasThiago: But just to make things clear, `group[key]` will never be `undefined` since this array is inittialised with `[]` 

https://github.com/vtex/toolbelt/blob/d7336a5c09c8c50f25438e27af9c79c2250513aa/src/oclif/hooks/init.ts#L235

dhasuda (approved):

PR #1068, opened 2021-04-09
Fix edition set command error handling

victorges:

This fixes 2 main issues with the `vtex edition set command`:
 - We were not handling responses to flow prompts, so we completely ignored when the user said
they didn't want to do some action like confirm the change in the `master` workspace or switching
to the sponsor account.
 - This is the more complex one which is almost a feature: to actually handle when the setEdition
API request actually returns a 404 error because of the `vtex.tenant-provisioner` app not being 
installed. Now we handle the corresponding not found errors from router and prompt the user to
already install the necessary app and retry the request after installing.

The 2 tricky parts were:
 - The route map of the platform has only eventual consistency, so we don't have guarantee anymore
that after installing an app, a call to some of its APIs will immediately work. So we implemented a retry
logic with an increasing delay to give some change to invalidate the route map cache and actually call
the app. 
 - The `appsInstall` command was actually using a static client so it didn't try to install apps in the
current context, but actually in the context where toolbelt started executing. Fixed that by creating 
some indirect functions that actually create a new client and then proxy the arguments to a function 
call in that client.

Broken flow of `vtex edition set` command, not respecting user rejections. Now no means no.

Also repeated errors with the `vtex.tenant-provisioner` app not being installed which cause 
unnecessary round-trips with the development team (🙋‍♂️ ) to fix the issue. Now toolbelt will already
detect the problem and fix it automatically.

 - `vtex edition set` and say no to some of the prompts, should stop the flow.
 - Uninstall the `vtex.tenant-provisioner` app from the sponsor and make sure toolbelt prompts to
install it and manages to set the edition afterwards.
 - Try to set an edition that doesn't exist and make sure toolbelt doesn't think it's the app that is not
installed.

Screen Shot 2021-04-09 at 17 53 11

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/src/api/modules/sponsor/setEdition.ts
+++ b/src/api/modules/sponsor/setEdition.ts
@@ -21,10 +21,22 @@ const promptSwitchToAccount = async (account: string, initial: boolean) => {
   return true
 }
 
+const confirmAndSwitchEnvironment = async (sponsorAccount: string, targetAccount: string, targetWorkspace: string) => {
+  if (!sponsorAccount) {
+    if (targetWorkspace !== 'master') {
+      throw createFlowIssueError('Can only set initial edition in master workspace')
+    }
+    return await promptSwitchToAccount('vtex', true)
+  }
+
+  const workspaceIsOk = targetWorkspace !== 'master' || await promptWorkspaceMaster(targetAccount)
+  return workspaceIsOk && await promptSwitchToAccount(sponsorAccount, false)
+}
+
 const tenantProvisionerApp = 'vtex.tenant-provisioner'
 
 const promptInstallTenantProvisioner = async (account: string) => {
-  log.notice(`Tenant provisioner app seems not to be installed in sponsor account ${chalk.blue(account)}.`)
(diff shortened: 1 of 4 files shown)


src/api/modules/sponsor/setEdition.ts
```diff
@@ -13,9 +15,83 @@ const promptSwitchToAccount = async (account: string, initial: boolean) => {
...
+    if (!isProvisionerNotInstalledError(err)) {
+      throw err
+    }
+    return false
+  }
+}
+
+const maxSetEditionRetries = 3
+
+const sleepSec = (sec: number) => new Promise(resolve => setTimeout(resolve, sec * 1000))
```
dhasuda, line 69: _non blocking_
This seems to be a very generic function. What do you think about moving it to another file, with more generic components? It could be a `utils` file for instance

dhasuda (approved):
LGTM!

gris:
@victorges thank you for the fix! May I ask you to open the same PR pointing to the `3.x` branch also? We are maintaining the two of them until 3.x gets stable.

PR #1019, opened 2021-02-12
Feature/app usd plan

marlonguimaraes:
#### What is the purpose of this pull request?

Currently all paid apps are only sold to Brazil and using the currency BRL (Brazilian Real). Our goal is to now release the App Store globally, selling apps to any country in USD (United State Dollars). This require some changes in the toolbelt which this PR attempts to solve.

#### What problem is this solving?

Handle `vtex install` for international apps.

#### How should this be manually tested?

There are multiple scenarios, first we have to make sure that old features still works and new ones work as expected.

_Brazilian VTEX account_

1) App is free: `vtex-test install vtex.social-selling` (This should work exactly as before)

image

2) Paid app, in BRL only: `vtex-test install vtex.sms-provider` (This should work exactly as before)

image

3) Paid app, available both in BRL and USD. User is prompted to choose which Plan.

`vtex-test install pilateslovers.billingoptionstest@20.x`

image

a) If the user chooses `BRL`, the install process should be exactly like the old one:

image

b) If the user chooses `USD` (or any thing other than `BRL`):

  i) If he account has an active contract (i.e that app was previously bought), the app should be automatically installed

  ii) Othwerwisea message is shown explaning that this app can only be installed and purchased in the App Store website

`vtex-test install pilateslovers.billingoptionstest@20.x`

image

Also, user is asked if she wants to go app's page in VTEX App Store. If Yes, a new tab is open in the user default browser.

4) App is only sold in USD 

`vtex-test install pilateslovers.billingoptionstest@81.x`

This should work exactly like 3) with option b) (i and ii)

Repeat all these tests with a international account like `vtexromania`.

#### Screenshots or example usage

Check above

#### Types of changes
- [ ] Refactor (non-breaking change that only makes the code better)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Requires change to documentation, which has been updated accordingly.

#### Chores checklist
- [x] Update `CHANGELOG.md`

PR #1081, opened 2021-04-22
Fix / isFreeApp

valentino-amadeus:
The purpose of this pull request is to classify an app as free during the installation process if the app declares no billing plan.

After some changes related to billing ([#1019](https://github.com/vtex/toolbelt/pull/1019)), users started to have problems when trying to install an app of type `sponsored` that didn't declare any billing plan. [Here](https://vtex.slack.com/archives/C2WSEU9QQ/p1618942741133300) we have an explanation of the behavior they have observed.

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/src/api/modules/utils.ts
+++ b/src/api/modules/utils.ts
@@ -135,8 +135,8 @@ export const appIdFromRegistry = (app: string, majorLocator: string) => {
 }
 
 const FREE_BILLING_OPTIONS_TYPE = 'free'
-export const isFreeApp = ({ type, free }: { type?: string; free?: boolean }) =>
-  type === FREE_BILLING_OPTIONS_TYPE || free
+export const isFreeApp = ({ type, free, plans }: { type?: string; free?: boolean; plans?: Plan[] }) =>
+  type === FREE_BILLING_OPTIONS_TYPE || free || !plans || plans.length === 0
 
 type BillingInfo = {
   subscription?: number
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- [install] Classify app as free if it declares no plan
+
 ## [2.127.1] - 2021-04-20
  - Fix `set edition` command to handle prompt cancellations
  - Add check on `set edition` command to install tenant-provisioner app in sponsor account


src/api/modules/utils.ts
```diff
@@ -135,8 +135,8 @@ export const appIdFromRegistry = (app: string, majorLocator: string) => {
 }
 
 const FREE_BILLING_OPTIONS_TYPE = 'free'
-export const isFreeApp = ({ type, free }: { type?: string; free?: boolean }) =>
-  type === FREE_BILLING_OPTIONS_TYPE || free
+export const isFreeApp = ({ type, free, plans }: { type?: string; free?: boolean; plans?: Plan[] }) =>
+  type === FREE_BILLING_OPTIONS_TYPE || free || !plans || plans.length === 0
```
gris, line 139:
```suggestion
  type === FREE_BILLING_OPTIONS_TYPE || SPONSORED_BILLING_OPTION_TYPE || free
```

gris (changes requested):

src/api/modules/utils.ts
gris: Please do the appropriate changes on constants and CHANGELOG as well.

valentino-amadeus: So it isn't (and it will never be) possible for an app of type `sponsored` to be not free?

gris: It is a good question. For that I will summon @marlonguimaraes and @victorges to help.

marlonguimaraes: I don't think it makes sense to make a sponsored paid app, if this exists I'm not aware. 

Fixing this inside `isFreeApp` doesn't sound semantic to mee. Maybe it should be in the place where `isFreeApp` evaluated, like:

```js
if(isFreeApp() || isSponsoredApp()) {
...
}
```

gris: Agreed

gris: Agreed

valentino-amadeus: Well, if there will indeed never be a paid sponsored app, I totally agree. But, however strange that may sound, are we absolutely certain that it will never happen? Do we really intend to avoid the existence of a paid sponsored app?

valentino-amadeus: The reason is that the same problem that we have with a `sponsored` app could happen with a `billable` app, _i.e._, we could have a `billable` app that doesn't declare billing plans.

gris: > so that any app that doesn't declare billing plans is considered free

Are we sure that this won't change in the future as well?

I think it is safer to test for equality of `type` instead of testing for absence of `plans`.

valentino-amadeus: Ok, I will close the PR, as there is no agreement on the value of its core purpose. I'll open a new PR to fix the bug the users have reported.

valentino-amadeus:
Closing because the proposed changes seem to be inappropriate. I'll fix the bug the users have reported in a different way, in a new PR.

src/api/modules/utils.ts
marlonguimaraes: So, we have an contract in the `billingOptions` API. It's up to the `billingOptions.type` inform the type of the app, so VTEX must ensure that contract holds and consumers to use the API accordingly. 

If for whatever reason anywhere we are not making that contract work, it's a bug and we must fix the root cause.

If the consumers/users are not following the API, we can send them the appropriate documentation or create/improve the documentation if it's not existent, provide support, improve our documentation etc . (Example: https://vtex.io/docs/recipes/development/setting-your-apps-billing-model/)

(IMHO)

PR #1082, opened 2021-04-22
[3.x] Fix / isFreeApp

valentino-amadeus:
The purpose of this pull request is to classify an app as free during the installation process if the app declares no billing plan.

This will keep consistency with the change introduced in #1081. 
Dealing with the case of a `sponsored` app that does not declare any billing plan also prevents similar problems to those that happened in the `2.x` version from happening here.

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


valentino-amadeus:
Closing because the proposed changes seem to be inappropriate. I'll fix the bug the users have reported in a different way, in a new PR.

PR #1090, opened 2021-05-10
Use templates as remote configs instead of hardcoded strings

williamads:
This pull request solves issue [#1073](https://github.com/vtex/toolbelt/issues/1073) by modifying the template code to use remote configuration instead of encoded strings.

By removing hardcodes, we can centralize system configs.

The expected result is the command to maintain its functioning. Running the command `vtex-test init`, we can see a list of options. 

This is normal operation
Captura de Tela 2021-05-10 às 11 46 24

#### Types of changes

#### Chores checklist
- [X] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+ - Use templates as remote configs instead of hardcoded strings.
+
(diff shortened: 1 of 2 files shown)


valentino-amadeus (approved):
LGTM

gris (approved):
LGMT. Usually I would suggest implementing the same behavior on `master` as well. But since we're going to drop that version soon and this is only a refactor, I wouldn't bother doing it.

PR #1110, opened 2021-07-21
Feature/adds release display-name flag

murilovarela:
This PR adds the `--display-name` flag to the `vtex release` command so the tag name is created with the app name from the manifest.

While developing in a monorepo tags can become a bit hard to understand what tag is related to what app, so adding a folder pattern to the tag name can solve this problem. Instead of `v0.10.1` we have `admin-ui@0.10.1`. 

#### Types of changes

#### Chores checklist
- [x] Update `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `--display-name` flag to release command
+
 ## [3.8.0-beta] - 2021-06-09
 
 ### Removed
--- a/src/modules/release/index.ts
+++ b/src/modules/release/index.ts
@@ -78,7 +78,7 @@ export default async (
     .split('/')
 
   // Pachamama v2 requires that version tags start with a 'v' character.
-  const tagText = `${displayName ? `${manifest.name}@` : ''}v${newVersion}`
+  const tagText = `${displayName ? `${manifest.name}@` : 'v'}${newVersion}`
   const changelogVersion = `\n\n## [${newVersion}] - ${year}-${month}-${day}`
 
   if (!(await utils.confirmRelease())) {
--- a/src/modules/release/index.ts
+++ b/src/modules/release/index.ts
@@ -78,7 +78,7 @@ export default async (
     .split('/')
 
   // Pachamama v2 requires that version tags start with a 'v' character.
-  const tagText = `${displayName ? manifest.name : ''}/v${newVersion}`
+  const tagText = `${displayName ? `${manifest.name}@` : ''}v${newVersion}`
   const changelogVersion = `\n\n## [${newVersion}] - ${year}-${month}-${day}`
 
   if (!(await utils.confirmRelease())) {
--- a/src/modules/release/index.ts
+++ b/src/modules/release/index.ts
@@ -4,6 +4,7 @@ import semver from 'semver'
 
 import log from '../../api/logger'
 import { ReleaseUtils } from './utils'
+import { ManifestEditor } from '../../api/manifest'
 
 export const releaseTypeAliases = {
   pre: 'prerelease',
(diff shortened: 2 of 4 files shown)


src/modules/release/index.ts
```diff
@@ -75,7 +78,7 @@ export default async (
     .split('/')
 
   // Pachamama v2 requires that version tags start with a 'v' character.
-  const tagText = `v${newVersion}`
+  const tagText = `${displayName ? manifest.name : ''}/v${newVersion}`
```
estacioneto, line 81: Have you thought about actually using `${manifest.name}@${newVersion}` to name the tag?

gris:
```suggestion
  const tagText = `${displayName ? manifest.name +"/": ''}v${newVersion}`
```

gris (changes requested):

src/modules/release/index.ts
murilovarela: Show! It is looking better with the `@` and follows our patter when deploying version `may-app@1.0.0` @estacioneto 
Also fixed the bug, @gris

estacioneto: Your code shows that the tag text is actually `app@v${version}`, or am I mistaken?

gris: Nice, agree with the suggestion to remove the `v` when the flag is set and also to update PR's description.

murilovarela: I forgot to push the change removing the `v`, my bad! 
Also updated the pr description!

gris (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `--display-name` flag to release command
```
estacioneto, line 12: (Requesting changes just to block the PR so this comment can be seen before the merge 😅 )

@gris, I've talked to @murilovarela and we thought about adding the vendor to the tag name. This might be useful to also quickly install an app just by copying the tag name. What do you think?

estacioneto (changes requested):

CHANGELOG.md
gris: If you guys think that it is a cool idea, I'm all in!
Just out of curiosity did you guys evaluated the usage of lerna, bazel or some other monorepo tooling?
I've just noticed that having a tag referring to a specific app inside a monorepo can be misleading, specially if you are updating more than 1 app at the same time (you will have to push 2 tags pointing to the same commit).

murilovarela: As I see we will have two commits with the version bump change, one for each app, and the tag will be pointing to that specific commit. 🤔

estacioneto: Can you talk about the work you're doing in the CI bot, @murilovarela?

gris: Hum, I see. I was just wondering if it makes more sense to have the tags point to versions of the monorepo itself, like this https://github.com/vtex/faststore/blob/master/lerna.json

estacioneto: In our case, @gris, it doesn't make sense to make all different apps have the same version since they don't need to be in sync to work. So we'd need to release a version of `vtex.admin-inventory` when `vtex.admin-logistics` changed, the way I saw.

We want to keep different tags so we can see, by the tag name, the app changed. I don't know if a `vtex.admin-logistics@{version}` tag with `vtex.logistics-carrier-graphql` changes would be a problem for us :)

gris: Nevermind, I've researched a little bit more and actually saw this process implemented with tags per package in other places. I'm very curious because I will implement a monorepo for my team in the future as well :D

murilovarela: The CI bot doesn't use the toolbelt. It generates the tag by itself. And what we decide here, I will replicate in the ci-hub code.

murilovarela: So we all agree that it is a good idea to add the vendor to the tag name, correct?

estacioneto: I see, @murilovarela. I meant to share knowledge about what we're doing since @gris is curious about the monorepo initiative.

Anyway, could you add the vendor in the tag name here too? 😁

gris: All good, it makes sense to have the vendor for me.

PR #1190, opened 2023-06-20
Make chrome debugger be opt-in instead of default

pedro823:
To make the chrome debugger tunnel be an explicit opt-in instead of a default open port.

This wastes resources on our cloud if the user is not actively using it. This allows us to intentionally measure the usage of this feature. 

Try linking an app normally and see that it doesn't start the chrome debugger server. By using `--with-debugger` in your linking flow, the previous behavior is restored.

#### Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to change)

#### Chores checklist
- [x] Update `CHANGELOG.md`


fabianaferreira (approved):
LGTM

PR #1235, opened 2024-10-03
Extend VBase clients with conflict handler endpoints

vitorflg:

- Extend the VBase client to have more options to handle conflicts

[Slack thread](https://vtex.slack.com/archives/C05CBNSEP0B/p1727825595441399)

- You can test it yarn linking all dependencies([@vtex/api](https://github.com/vtex/node-vtex-api/pull/561), this module, and the [cli-plugin-workspace](https://github.com/vtex/cli-plugin-workspace/pull/14)), and running with the alias:
vtex-test promote After that, listing to conflicts should return an empty array.

#### Types of changes

* [ ] Bug fix (a non-breaking change which fixes an issue)
* [X] New feature (a non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Requires change to documentation, which has been updated accordingly.

### DIFF
--- a/src/api/clients/IOClients/infra/VBase.ts
+++ b/src/api/clients/IOClients/infra/VBase.ts
@@ -22,6 +22,7 @@ export class VBase extends InfraClient {
     })
   }
 
+  // Resolve a specific pages-graphql conflict
   public resolveConflict = (bucketKey: string, path: string, content: any) => {
     const data = [
       {
@@ -36,22 +37,17 @@ export class VBase extends InfraClient {
     })
   }
 
+  // List all conflicts in the pages-graphql bucket
   public getConflicts = async () => {
     return this.http.get(routes.Conflicts('vtex.pages-graphql/userData'), {
       metric: 'vbase-get-conflicts',
     })
   }
 
+  // Verify if there is at least one conlfict in the pages-graphql bucket
   public checkForConflicts = async () => {
-    let status: number
-    try {
-      const response = await this.http.get(routes.File('vtex.pages-graphql/userData', 'store/content.json'), {
-        metric: 'vbase-conflict',
-      })
-      status = response.status
-    } catch (error) {
-      status = error.response && error.response.status
-    }
-    return status === 409
+    const response = await this.getConflicts().then(res => res.data)
+
+    return response?.data?.length > 0
   }
 }
--- a/src/api/clients/IOClients/infra/VBase.ts
+++ b/src/api/clients/IOClients/infra/VBase.ts
@@ -2,8 +2,9 @@ import { InfraClient, InstanceOptions, IOContext } from '@vtex/api'
 import { IOClientFactory } from '../IOClientFactory'
 
 const routes = {
-  Bucket: (bucket: string) => `/buckets/${bucket}`,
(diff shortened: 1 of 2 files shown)


src/api/clients/IOClients/infra/VBase.ts
```diff
@@ -21,10 +22,30 @@ export class VBase extends InfraClient {
...
+    return this.http.get(routes.Conflicts('vtex.pages-graphql/userData'), {
+      metric: 'vbase-get-conflicts',
+    })
+  }
+
   public checkForConflicts = async () => {
     let status: number
     try {
-      const response = await this.http.get(routes.File('vtex.pages-graphql', 'store/content.json'), {
+      const response = await this.http.get(routes.File('vtex.pages-graphql/userData', 'store/content.json'), {
```
victorhmp, line 48: Should this also be extended to accept a `filePath` parameter? To support the template specific files.

victorhmp (approved):

src/api/clients/IOClients/infra/VBase.ts
vitorflg: not necessary. This function is for resolving the conflict just for one file. It's not even required we keep this logic, but I decided to let it to avoid side effects

mateuspontes (approved):

src/api/clients/IOClients/infra/VBase.ts
```diff
@@ -21,10 +22,30 @@ export class VBase extends InfraClient {
     })
   }
 
+  public resolveConflict = (bucketKey: string, path: string, content: any) => {
```
filafb, line 25: Nit: WDYT about adding jsdoc annotation in these new functions? Since they are used in a outside context, it might be helpful to add some shared comments. I think it's important to mention they are very specific for pages-graphql stuff.

filafb (changes requested):
The changed required here is just the  target branch, which should be `main`, not `2.x`.

This VBase client is very specific for managing buckets and files for pages-graphql. In the new content structure, will this client still be needed or changed in any way?

src/api/clients/IOClients/infra/VBase.ts
vitorflg: Agree! But I wonder if it's only used for pages-graphql.

src/api/clients/IOClients/infra/VBase.ts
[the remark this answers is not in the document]
vitorflg: I will extend it and put the added logic inside this checkForConflicts condition.

src/api/clients/IOClients/infra/VBase.ts
filafb: I think it's safe to assume that since the bucket name has pages-graphql hard coded in the path.

vitorflg:
> The changed required here is just the target branch, which should be `main`, not `2.x`.
> 
> This VBase client is very specific for managing buckets and files for pages-graphql. In the new content structure, will this client still be needed or changed in any way?

It will be needed and we won't have to change it! I created another PR with main as base: https://github.com/vtex/toolbelt/pull/1236

PR #1236, opened 2024-10-11
Extend/vbase client v2

vitorflg:

- Extend the VBase client to have more options to handle conflicts

[Slack thread](https://vtex.slack.com/archives/C05CBNSEP0B/p1727825595441399)

- You can test it yarn linking all dependencies([@vtex/api](https://github.com/vtex/node-vtex-api/pull/561), this module, and the [cli-plugin-workspace](https://github.com/vtex/cli-plugin-workspace/pull/14)), and running with the alias:
vtex-test promote After that, listing to conflicts should return an empty array.

#### Types of changes

* [ ] Bug fix (a non-breaking change which fixes an issue)
* [X] New feature (a non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Requires change to documentation, which has been updated accordingly.

### DIFF
--- a/src/api/clients/IOClients/infra/VBase.ts
+++ b/src/api/clients/IOClients/infra/VBase.ts
@@ -48,6 +48,6 @@ export class VBase extends InfraClient {
   public checkForConflicts = async () => {
     const response = await this.getConflicts().then(res => res.data)
 
-    return response?.data?.length > 0
+    return response?.length > 0
   }
 }
--- a/src/api/clients/IOClients/infra/VBase.ts
+++ b/src/api/clients/IOClients/infra/VBase.ts
@@ -22,6 +22,7 @@ export class VBase extends InfraClient {
     })
   }
 
+  // Resolve a specific pages-graphql conflict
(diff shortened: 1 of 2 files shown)


src/api/clients/IOClients/infra/VBase.ts
```diff
@@ -21,16 +22,32 @@ export class VBase extends InfraClient {
...
-        metric: 'vbase-conflict',
-      })
-      status = response.status
-    } catch (error) {
-      status = error.response && error.response.status
-    }
-    return status === 409
+    const response = await this.getConflicts().then(res => res.data)
+
+    return response?.data?.length > 0
```
filafb, line 51: If I am not wrong, axios throws an error for 4xx responses. If that is the case, I think we are changing the behavior here, since the line 49 will raise the error and will not get to the return statement. Have you checked that? Also, as in the other PR, I would avoid the `await / .then` combination.

mateuspontes (approved):

PR #1248, opened 2025-06-02
Update vtexid host to use myvtex

iagoaraujo:
Use `myvtex.com` instead of `vtex.com.br` on VTEX ID client.

Currently the toolbelt uses the VTEX via the path `vtexid.vtex.com.br`, this is a deprecated URL that we need to decommission and also there the work related to the mTLS implementation, where requests to `vtex.com.br` will be blocked.

Run `yarn watch` and try to login:

- Accounts with mTLS disabled and VPN Full Tunnel off. Ex: vtexgame2
- Accounts with mTLS disabled and VPN Full Tunnel on. Ex: vtexgame2
- Accounts with mTLS enabled and VPN Full Tunnel off. Ex: ismarketplacedes
-- On this scenario you should receive an erros, because in case of mTLS enabled the user should be using the VPN.
- Accounts with mTLS enabled and VPN Full Tunnel on. Ex: ismarketplacedes

#### Types of changes

#### Chores checklist
- [ ] Update `CHANGELOG.md`

### DIFF
--- a/src/api/clients/IOClients/external/VTEXID.ts
+++ b/src/api/clients/IOClients/external/VTEXID.ts
@@ -7,7 +7,6 @@ import { switchOpen } from '../../../../modules/featureFlag/featureFlagDecider'
 export class VTEXID extends IOClient {
   private static readonly DEFAULT_TIMEOUT = 10000
   private static readonly DEFAULT_RETRIES = 2
-  private static readonly BASE_URL = 'https://vtexid.vtex.com.br'
   private static readonly API_PATH_PREFIX = '/api/vtexid'
   private static readonly TOOLBELT_API_PATH_PREFIX = `${VTEXID.API_PATH_PREFIX}/toolbelt`
   private static readonly VTEX_ID_AUTH_COOKIE = 'VtexIdClientAutCookie'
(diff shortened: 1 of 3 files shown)


filafb (approved):

src/api/clients/IOClients/external/VTEXID.ts
```diff
@@ -28,27 +27,27 @@ export class VTEXID extends IOClient {
       timeout: VTEXID.DEFAULT_TIMEOUT,
       retries: VTEXID.DEFAULT_RETRIES,
       ...opts,
-      baseURL: VTEXID.BASE_URL,
```
arturpimentel, line 31: Will this work in a first-login/account change scenario? My fear is that `ioContext.account` could be `undefined` or wrong when this client is called

iagoaraujo: @arturpimentel the first-login will always ask for an account. Account switch it is also mandatory to specify the account otherwise it throws an error
image
image

arturpimentel: Nice! Thanks for checking that!

arturpimentel (approved):
LGTM

PR #1270, opened 2026-06-24
docs: spec for removing login dependency on help command [XTNSNS-2080]

silvadenisaraujo:
## What

Adds a spec under `specs/help-without-login.md` describing how to remove the login dependency when running `help` in the CLI, so agents (and users) can discover commands without credentials.

## Why

`src/oclif/hooks/init.ts` forces an interactive login for any command not in an `allowedList`. Running `vtex <command> --help` (e.g. `vtex deploy --help`) triggers a login prompt even though only static help text is requested, blocking command discovery on logged-out machines.

## Spec summary

- Add a pure helper `isHelpInvocation(commandId, argv)` in `src/oclif/hooks/utils.ts` that detects all help forms (`help` command and `--help`/`-h` flags), stopping at the `--` separator.
- Wire it into `checkLogin` to skip login for help invocations.
- Add table-driven unit tests in `src/oclif/hooks/utils.test.ts`.
- Help rendering has no network/auth dependency (FeatureFlag reads a local Configstore file).

## Tracked follow-up (deferred, out of scope)

`vtex help` can crash on a clean machine with an empty feature-flag cache (`Object.keys(undefined)` in `showHelp`). Proposed fix: fall back to a single default group when `COMMANDS_GROUP`/`COMMANDS_GROUP_ID` are missing. Tracked in the spec.

This PR contains the **spec only**; implementation follows separately.

### DIFF
--- a/specs/help-without-login.md
+++ b/specs/help-without-login.md
@@ -1,0 +1,76 @@
+# Spec: Remove login dependency for `help` in the toolbelt CLI
+
+## Problem
+
+`src/oclif/hooks/init.ts` → `main()` always calls `checkLogin(options.id)`, which
+forces an interactive login for any command not present in an `allowedList`.
+Running `vtex <command> --help` (e.g. `vtex deploy --help`) sets
+`options.id = 'deploy'` (not in the allowed list) and triggers a login prompt,
+even though only static help text is requested. This blocks agents (and users)
+from discovering commands without credentials.
+
+## Goal
+
+All help invocations render without requiring login, so agents on any machine can
+discover the proper commands to run.
+
+## Scope
+
+**In scope** — skip login for every help form:
+
+- `vtex help`
+- `vtex help <command>`
+- `vtex <command> --help` / `-h`
+- `vtex --help` / `-h`
+- `vtex` (no args — already allowed via `undefined`)
+
+**Out of scope** — strictly the login gate only. See the tracked follow-up issue
+below for the empty-feature-flag-cache crash, which is intentionally deferred.
+
+## Design decisions
+
+1. **Generic help detection, not per-command whitelisting.** Whitelisting each
+   command cannot cover `vtex deploy --help`, so we detect "this is a help
+   invocation" generically.
+
+2. **New helper `isHelpInvocation(commandId, argv)` in
+   `src/oclif/hooks/utils.ts`** (next to `getHelpSubject`). Returns `true` when:
+   - `commandId === 'help'`, or
+   - `argv` contains `--help` or `-h` **before** any `--` separator.
+   - `-h` / `--help` are safe to treat as help: `-h` is globally reserved for
+     help in `src/api/oclif/CustomCommand.ts`, with no other usage.
+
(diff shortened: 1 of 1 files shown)


specs/help-without-login.md
```diff
@@ -0,0 +1,76 @@
+# Spec: Remove login dependency for `help` in the toolbelt CLI
```
wisneycardeal, line 1: The spec is well-scoped and correct: adding a small pure helper isHelpInvocation(argv, commandId) and skipping the login gate for help invocations is the right, low-risk way to enable help rendering without credentials.


Two implementation notes/improvements are recommended while landing the spec:
- Make checkLogin accept argv (or use process.argv) and early-return when isHelpInvocation(...) is true.

- Defensively handle missing/undefined feature-flag values in showHelp to avoid the tracked crash (Object.keys on undefined). I recommend a small fallback so help always renders even if the feature-flag cache is empty.


Possible improvements (actionable):

- Include the defensive feature-flag fallback in the same PR. It's a single-line hardening change and avoids the separate crash the spec calls out.

- Add an integration smoke test (shell test) that runs node bin/run help or similar in an environment that simulates no credentials and ensures no interactive prompt is emitted. If that is heavy, at least add the unit tests for isHelpInvocation and a unit test that checkLogin returns early for help argv (mock SessionManager.getSingleton().checkValidCredentials to false and assert authLogin not called).

- Add a short comment in utils.ts documenting the separation handling to avoid future regressions.

silvadenisaraujo: Thanks for the review! All four suggestions were incorporated into the spec in eeb58119:

- **argv wiring**: `checkLogin` now takes an explicit `argv` parameter (`checkLogin(command, argv)`) and early-returns when `isHelpInvocation(command, argv)` is true — using `options.argv` from the init hook rather than `process.argv`, for testability (design decision 4).
- **Feature-flag fallback**: pulled into scope for the same PR. `showHelp` will fall back to rendering all commands under a single default group when `COMMANDS_GROUP` / `COMMANDS_GROUP_ID` are missing (design decision 6).
- **Tests**: added a unit test that `checkLogin` early-returns for help argv (mocking `SessionManager`, asserting `authLogin` is not called) plus an integration smoke test running `node bin/run help` with no credentials and an empty feature-flag cache — which also exercises the fallback.
- **Docs**: a short comment in `utils.ts` documenting the `--` separator handling (design decision 7).

src/oclif/hooks/utils.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    ['-h', [], true],
+    [undefined, ['--help'], true],
+    [undefined, ['-h'], true],
+    ['workspace', ['use', '--help'], true],
+    // non-help forms → false
+    ['deploy', [], false],
+    ['deploy', ['--', '--help'], false],
+    ['deploy', ['--', '-h'], false],
+    [undefined, [], false],
+    ['link', ['--verbose'], false],
```
juliobguedes, line 22:
```suggestion
    ['link', ['--verbose'], false],
    ['login', '-h'],
    ['login', '--help']
```

